PR #8, opened 2015-09-08
Adds docker-compose instructions

adriancole:
cc @dsyer @spencergibb @abesto


adriancole:
Note: the docker image is 103.5 MB, as it layers only a little on top of zipkin-base

```
openzipkin/zipkin-java-server         latest              ac388a9b54b1        12 minutes ago      103.5 MB
```

PR #10, opened 2015-09-20
Adds WIP JDBC backend

adriancole:
Adds an experimental MySQL backend which uses a simpler table structure
than upstream. It is simpler as it keeps a single table for annotations
regardless of whether they are "binary" or not.

While not yet hooked up, this passes all tests in SpanStoreSpec, and I
also succesfully deployed it to EC2, targetting an RDS MySQL instance.


adriancole:
cc @dsyer @spencergibb @jfeltesse-mdsol I'm inclined to merge and progress eventhough I've not woven in the SpanStoreSpec nicely (and/or have the impl testing against an in-memory or travis based mysql, yet). thoughts?

dsyer:
You mean the tests pass but not yet automated? Sounds good to me.

adriancole:
Exactly. I ran them manually as they need to point to a MySQL instance. I
can follow-up with a Travis related change that runs them automatically.

spencergibb:
:+1:

PR #13, opened 2015-09-23
Adds integration tests from scala project and addresses drift

adriancole:
Running the same tests as the scala side of OpenZipkin should help increase confidence in the java port.

### DIFF
--- a/zipkin-java-scribe/src/main/java/io/zipkin/scribe/ScribeSpanConsumer.java
+++ b/zipkin-java-scribe/src/main/java/io/zipkin/scribe/ScribeSpanConsumer.java
@@ -26,10 +26,10 @@ import org.apache.thrift.transport.TMemoryBuffer;
 
 public final class ScribeSpanConsumer implements Scribe {
 
-  private final Consumer<Iterable<Span>> consumer;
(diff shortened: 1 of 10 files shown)


zipkin-java-scribe/src/main/java/io/zipkin/scribe/ScribeSpanConsumer.java
```diff
@@ -48,8 +48,7 @@ public ResultCode log(List<LogEntry> messages) {
             return null;
           }
         })
-        .filter(s -> s != null)
-        .filter(s -> !(s.isClientSide() && s.serviceNames().contains("client")));
```
adriancole: this is a goofy special case I'll remove from upstream.

adriancole:
ok.. now to try and see if the mysql passes against this

zipkin-java-scribe/src/main/java/io/zipkin/scribe/ScribeSpanConsumer.java
adriancole: https://github.com/openzipkin/zipkin/pull/724

PR #14, opened 2015-09-23
Renames zipkin-query to zipkin-spanstore

adriancole:
This replaces `ZipkinQuery & Consumer<List<Span>>` with `SpanStore`,
simplifying type signatures and exposing the storage layer in its
entirety.

In the scala project, there's a SpanStore contract, a read/write api
similar to a union of the Scribe and ZipkinQuery interfaces. The java
project expects scribe to fade into an implementation detail. However,
there's still value in encapsulating the storage aspect, including
writes, as that's how we can eventually replace the storage layer in
any listener. It also helps us to write tests.

The down-side is that in many cases, the code to store is far simpler
than the code to query. Joining the two apis means you need to bring
along both when you want either. I consider this a minor concern, as
the code is still reasonably small and would have the same dependencies
either way.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
   <modules>
     <module>swift-codec</module>
     <module>zipkin-java-core</module>
-    <module>zipkin-java-query</module>
+    <module>zipkin-java-spanstore</module>
     <module>zipkin-java-scribe</module>
     <module>zipkin-java-dependencies</module>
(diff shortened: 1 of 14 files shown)


adriancole:
cc @kristofa @dsyer @spencergibb

spencergibb:
:+1:

PR #15, opened 2015-09-24
Updates to zipkin 1.10.0 and adds MySQL tests

adriancole:


adriancole:
@dsyer @spencergibb @kristofa so this conditionally runs the mysql tests when in travis. These are literally the same tests as the scala side until I make the java ones match behavior exactly.

zipkin-java-jdbc/src/test/java/io/zipkin/jdbc/JDBCScalaSpanStoreTest.java
```diff
@@ -0,0 +1,41 @@
...
+import io.zipkin.spanstore.ScalaSpanStoreAdapter;
+import java.sql.SQLException;
+import org.junit.BeforeClass;
+
+public class JDBCScalaSpanStoreTest extends SpanStoreSpec {
+  private static JDBCSpanStore spanStore;
+
+  @BeforeClass
+  public static void setupDB() {
+    spanStore = new JDBCTestGraph().spanStore;
```
adriancole: when there's no MYSQL_ env, this throws assumption violated, which will circuit-break the tests into ignore state.

.travis.yml
```diff
@@ -9,5 +9,13 @@ language: java
 jdk:
   - oraclejdk8
 
+services:
+  - mysql
+
+before_install:
```
adriancole: so cool we can now test our readme instructions

PR #16, opened 2015-09-30
Packages dependency-free core with json and thrift codecs

adriancole:
### New dependency-free core jar with built-in thrift and json
This removes all dependencies on FB swift for thrift, and instead adds
`Codec.THRIFT`. This internally uses libthrift.

This adds `Codec.JSON`, which is compatible with the query api from
the scala project. This internally uses moshi.

The result is a <250Kb core jar with no dependencies, which works with
java 7. This makes it possible to use the core jar upstream or in 
other projects which need to parse spans or dependencies.

### Background 
FB swift didn't play nice with auto-value, and also includes heaps of 
dependencies. That would be ok, but the project is on life-support, with
no activity since June. This pushed me into figuring out a different 
way to support thrift and json. 

One of the hardest parts was how to organize the code so that it could
run interop tests against the scala codebase. The vastness of upstream's
dependency graph confuses shade, and adds a lot of labor to maven. 
Instead, I made an interop project, which can import all the things that
need to be tested from the scala project


spencergibb:
:+1:

adriancole:
update:

code's a bit dirty, but made some significant progress:

* both the in-memory and mysql storage backends pass all the zipkin storage tests (even the weird ones)
* the json model is compatible with the one in the scala project
* next step is an http straw-man that responds the the query api (and yank remaining thrift RPC stuff)

PR #17, opened 2015-10-05
Adds HTTP Java Service

adriancole:
This adds an implementation of the Zipkin HTTP Api, using Spring Boot.

Further work can be added later, such as tracing and stats related to
zipkin operations.

Tested locally and on Docker

PR #18, opened 2015-10-07
Make it more Bootiful

dsyer:
Some small changes that make the server app more 
idiomatically Spring Boot.

### DIFF
--- a/zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
+++ b/zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
@@ -13,6 +13,10 @@
  */
 package io.zipkin.interop;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.twitter.util.Future;
@@ -20,12 +24,10 @@ import com.twitter.zipkin.common.Span;
 import com.twitter.zipkin.json.JsonSpan;
 import com.twitter.zipkin.json.ZipkinJson$;
 import com.twitter.zipkin.storage.QueryRequest;
+
 import io.zipkin.Codec;
 import io.zipkin.SpanStore;
 import io.zipkin.internal.Nullable;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
 import scala.Tuple2;
 import scala.collection.Iterator;
 import scala.collection.JavaConversions;
@@ -63,14 +65,14 @@ public final class ScalaSpanStoreAdapter extends com.twitter.zipkin.storage.Span
       Tuple2<String, String> keyValue = i.next();
       request.addBinaryAnnotation(keyValue._1(), keyValue._2());
     }
-    return toSeqFuture(spanStore.getTraces(request.build()));
+    return toSeqFuture(this.spanStore.getTraces(request.build()));
   }
 
   @Override
   public Future<Seq<Seq<Span>>> getTracesByIds(Seq<Object> input) {
     List<Long> traceIds = new ArrayList<>(input.size());
-    for (Iterator<Object> i = input.iterator(); i.hasNext(); traceIds.add((Long) i.next())) ;
-    return toSeqFuture(spanStore.getTracesByIds(traceIds));
+    for (Object i : JavaConversions.asJavaCollection(input)) { traceIds.add((Long) i);}
+    return toSeqFuture(this.spanStore.getTracesByIds(traceIds));
   }
 
   static Future<Seq<Seq<Span>>> toSeqFuture(List<List<io.zipkin.Span>> traces) {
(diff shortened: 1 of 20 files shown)


zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
```diff
@@ -63,14 +65,14 @@ public ScalaSpanStoreAdapter(SpanStore spanStore) {
...
-    return toSeqFuture(spanStore.getTraces(request.build()));
+    return toSeqFuture(this.spanStore.getTraces(request.build()));
   }
 
   @Override
   public Future<Seq<Seq<Span>>> getTracesByIds(Seq<Object> input) {
     List<Long> traceIds = new ArrayList<>(input.size());
-    for (Iterator<Object> i = input.iterator(); i.hasNext(); traceIds.add((Long) i.next())) ;
-    return toSeqFuture(spanStore.getTracesByIds(traceIds));
+    for (Object i : JavaConversions.asJavaCollection(input)) { traceIds.add((Long) i);}
```
eirslett: How about
```
List<Long> traceIds = JavaConversions.asJavaCollection(input).stream().map(Long::valueOf).collect(toList())
```

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
```diff
@@ -85,35 +87,35 @@ public ScalaSpanStoreAdapter(SpanStore spanStore) {
...
-    return Future.value(JavaConversions.asScalaBuffer(spanStore.getSpanNames(service)).toSeq());
+    return Future.value(JavaConversions.asScalaBuffer(this.spanStore.getSpanNames(service)).seq());
   }
 
   @Override
   public Future<BoxedUnit> apply(Seq<Span> input) {
     List<io.zipkin.Span> spans = new ArrayList<>(input.size());
-    for (Iterator<Span> i = input.iterator(); i.hasNext(); ) {
-      io.zipkin.Span span = invert(i.next());
+    for (Span i : JavaConversions.asJavaCollection(input)) {
```
eirslett: and maybe `map(this::invert)` here? (Or whatever the Java syntax is)

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: because the core is java 7

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: same reason. while servers should be able to be made in java 8, this isn't good enough reason to make the core model java 8.

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: oh wait.. I forgot I moved this to the interop package! I suppose we can revert to this (was like that in an earlier rev)

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: disregard my comment.

zipkin-java-server/src/main/java/io/zipkin/server/EnableZipkinServer.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.context.annotation.Import;
+
+/**
+ * @author Dave Syer
```
adriancole: no author tags por favor :)

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerConfiguration.java
```diff
@@ -11,53 +11,43 @@
...
-      String url = String.format("jdbc:mysql://%s:%s/zipkin?user=%s&password=%s&autoReconnect=true",
-          mysqlHost, mysqlPort, mysqlUser, mysqlPass);
-
-      HikariDataSource datasource = new HikariDataSource();
-      datasource.setDriverClassName("com.mysql.jdbc.Driver");
-      datasource.setJdbcUrl(url);
-      datasource.setMaximumPoolSize(10);
-      datasource.setConnectionTestQuery("SELECT '1'");
+  SpanStore provideSpanStore(DataSource datasource) {
+    if (datasource !=null && this.server.getStore().getType()==Type.jdbc) {
```
adriancole: neat!

zipkin-java-server/src/main/resources/zipkin-server.yml
```diff
@@ -0,0 +1,26 @@
...
+  password: ${MYSQL_PASS:root}
+  db: ${MYSQL_DB:zipkin}
+zipkin:
+  collector:
+    port: ${COLLECTOR_PORT:9410}
+  store:
+#    type: jdbc # default is inMemory
+logging:
+  level:
+# Silence Invalid method name: '__can__finagle__trace__v3__'
```
adriancole: :love_letter:

zipkin-java-server/pom.xml
```diff
@@ -100,13 +117,9 @@
...
         <artifactId>spring-boot-maven-plugin</artifactId>
-        <executions>
-          <execution>
-            <goals>
-              <goal>repackage</goal>
-            </goals>
-          </execution>
-        </executions>
+        <configuration>
+          <classifier>exec</classifier>
```
adriancole: does this invalidate the docker instructions in the README? or does exec imply repackage? if the former, I suppose we just need to change the instructions in the README?

zipkin-java-server/pom.xml
```diff
@@ -69,17 +83,20 @@
...
+      <optional>true</optional>
     </dependency>
 
     <dependency>
       <groupId>mysql</groupId>
       <artifactId>mysql-connector-java</artifactId>
+      <optional>true</optional>
     </dependency>
 
     <dependency>
```
adriancole: drop this dep as the classes are no longer loaded

zipkin-java-server/pom.xml
adriancole: disregard I think.. :)

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: done'ish. `Seq(Object)` sucks

zipkin-java-interop/src/main/java/io/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: done via d162ef3444952ca9d00c05a8c776021dbe08695b

adriancole:
@dsyer wrt "Aside: wouldn't 'FROM java:8' work just as well?" This would make our base layer inflate back up to I think 400MB. This increases the time for the docker publish to run, which is already quite long, and makes working on bad wifi worse.

adriancole:
picked first commit into master, shuffled others into #17

PR #20, opened 2015-10-07
Trim dependencies a bit

dsyer:
The main problem     is that there is a clash between 2 impls
of jsr-303. One of them has to be exlcuded.


spencergibb:
:+1:

PR #21, opened 2015-10-08
Fix JOOQ config so schema is not rendered by default

dsyer:


adriancole:
Nice one. I did this in the tests, but wasn't sure if we should inject the settings or hardcode this long term. Regardless, this is a good idea.

PR #23, opened 2015-10-09
Adds tracing for Api and JDBC layers

adriancole:
This adds a Brave tracer, adding Server Receive and Server Send annotations for `zipkin-query`.

This uses JOOQ to add to Client Send and Client Receive traces for the service named `zipkin-jdbc`


adriancole:
cc @dsyer @spencergibb @kristofa

adriancole:
Here's the trace depth now

screen shot 2015-10-09 at 1 24 30 pm

adriancole:
Here's an example showing the actual SQL query

screen shot 2015-10-09 at 1 29 15 pm

kristofa:
Nice! We could add JOOQ support (with read and write) to brave.

adriancole:
@kristofa yep especially now that zipkin-java no longer does any writes to the database!

adriancole:
oh wait.. it does, but yeah in general, it would be a cool feature, and quite easy to configure

PR #24, opened 2015-10-09
Refactor to avoid using package private stuff in Brave

dsyer:
Add support for MySQL tracing (broken because of cycle)
Move static beans into private local methods

### DIFF
--- a/zipkin-java-server/src/main/java/io/zipkin/server/brave/ApiTracerConfiguration.java
+++ b/zipkin-java-server/src/main/java/io/zipkin/server/brave/ApiTracerConfiguration.java
@@ -12,22 +12,22 @@
  * the License.
  */
 // TODO: switch package back after https://github.com/openzipkin/brave/pull/99
-package com.github.kristofa.brave;
-
-import java.net.Inet4Address;
-import java.net.InetAddress;
-import java.net.NetworkInterface;
-import java.net.SocketException;
-import java.security.SecureRandom;
-import java.util.Collections;
+package io.zipkin.server.brave;
 
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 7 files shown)


zipkin-java-server/src/main/java/io/zipkin/server/brave/ApiTracerConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+@Configuration
+@Import(MySqlTracerConfiguration.class)
+public class ApiTracerConfiguration extends WebMvcConfigurerAdapter {
+
+  @Value("${zipkin.collector.port:9410}")
+  int scribePort;
+
+  @Bean
+  Brave brave() {
+    return new Brave.Builder("zipkin-query").spanCollector(braveSpanCollector()).build();
```
adriancole: so much nicer

zipkin-java-server/src/main/java/io/zipkin/server/brave/MySqlTracerConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.context.annotation.Configuration;
+
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.mysql.MySQLStatementInterceptor;
+import com.mysql.jdbc.Driver;
+
+/**
+ * @author Dave Syer
```
adriancole: if you're up, can you scrub this, otherwise I can

dsyer: Oops. I need to remember to switch those off. Spring sources always have them.

dsyer:
Last commit removed author tag.

PR #28, opened 2015-10-10
Adds on-demand dependencies support to zipkin-jdbc

adriancole:
This allows you to see the dependency-graph without hadoop or otherwise.

This shows services which are linked by parentId, with a count of how
many spans were logged. This will only show traces with at least one
edge (child), even if that edge is a loopback.

The UI remains unchanged: the point of the line corresponds with the amount of traffic.

screen shot 2015-10-09 at 10 00 33 pm


adriancole:
here's the corresponding functionality in the scala project https://github.com/openzipkin/zipkin/pull/769

PR #29, opened 2015-10-10
Prototype Brave SpanCollector using SpanStore

dsyer:
With this change the thrift server is not used by Brave (it sends spans directly top the span store). This only affects the tracing of the requests to the query service.

### DIFF
--- a/zipkin-java-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-java-server/src/main/resources/zipkin-server.yml
@@ -18,6 +18,7 @@ server:
 spring:
   datasource:
     driver-class-name: com.mysql.jdbc.Driver
+# add this for mysql tracing: &statementInterceptors=com.github.kristofa.brave.mysql.MySQLStatementInterceptor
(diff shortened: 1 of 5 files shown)


adriancole:
so love this!

zipkin-java-server/src/main/resources/zipkin-server.yml
```diff
@@ -18,6 +18,7 @@ server:
 spring:
   datasource:
     driver-class-name: com.mysql.jdbc.Driver
+# add this for mysql tracing: &statementInterceptors=com.github.kristofa.brave.mysql.MySQLStatementInterceptor
```
adriancole: do we want to suggest this? we're already tracing sql statements.

kristofa:
Nice!

adriancole:
I'm going to pick this in, just going to format the files and scrub the redundant tracing suggestion (since we already add annotations that the mysql tracer would)

adriancole:
closed via ab9012af3120a340cd49f9daa998d14525be91ba thanks so much!

PR #31, opened 2015-10-11
Kills dead code related to unused Dependencies type

adriancole:
This type is no longer used, so totally worth scrapping


adriancole:
this implicitly simplified codec, as it only has to deal with two types now.

PR #32, opened 2015-10-12
Add snapshot publishing

openzipkin-release:
* publish to sonatype if merge is not a PR
* added the gpg block in case we need it.  For some reason I left passphrase out and it still worked O_o.  Checking in keys also seemed some what janky.  Please let me know if you have an opinion on this.
(might be the maven-gpg-plugin thats installed)
* openzipkin-release (openzipkin@gmail.com) user
* need to do a test when code is merged into master
* TRAVIS_BRANCH - is the branch you are merging into, which almost always is master.  TRAVIS_PULL_REQUEST - is either false or the PR number.  This makes the assumption that if you are on master and it is not a pull request, then the result is something to snapshot.

### DIFF
--- a/.settings.xml
+++ b/.settings.xml
@@ -1,0 +1,42 @@
+<!--
+
+    Copyright 2015 The OpenZipkin Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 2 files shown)


adriancole:
Nice start! Looks like build fail as it tried to publish the PR

PR #33, opened 2015-10-13
Tidies up docker configuration with upstream

adriancole:
Makes the docker side of the business just like the others

### DIFF
--- a/zipkin-java-server/Dockerfile
+++ b/zipkin-java-server/Dockerfile
@@ -12,16 +12,16 @@
 # the License.
 #
 
-FROM openzipkin/zipkin-base:base-1.17.0
+FROM openzipkin/zipkin-base:base-1.17.1
 
 MAINTAINER OpenZipkin "http://zipkin.io/"
 
 VOLUME /tmp
 
-ADD target/zipkin-java-server-0.1.0-SNAPSHOT-exec.jar /usr/local/bin/zipkin-server.jar
+ADD target/zipkin-java-server-0.1.0-SNAPSHOT-exec.jar zipkin-server.jar
 
-RUN sh -c 'touch /usr/local/bin/zipkin-server.jar'
+RUN touch zipkin-server.jar
 
-EXPOSE 9410 9411
+EXPOSE 9411
 
-ENTRYPOINT ["java","-Djava.security.egd=file:/dev/./urandom","-jar","/usr/local/bin/zipkin-server.jar"]
+CMD test -n "$STORAGE_TYPE" && source .${STORAGE_TYPE}_profile; java -Djava.security.egd=file:/dev/./urandom -jar zipkin-server.jar
--- a/zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerProperties.java
+++ b/zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerProperties.java
@@ -30,10 +30,10 @@ class ZipkinServerProperties {
 
   static class Store {
     enum Type {
-      jdbc, inMemory;
+      mysql, mem;
     }
 
-    private Type type = Type.inMemory;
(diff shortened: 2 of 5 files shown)


zipkin-java-server/Dockerfile
```diff
@@ -12,16 +12,16 @@
...
+ADD target/zipkin-java-server-0.1.0-SNAPSHOT-exec.jar zipkin-server.jar
 
-RUN sh -c 'touch /usr/local/bin/zipkin-server.jar'
+RUN touch zipkin-server.jar
 
-EXPOSE 9410 9411
+EXPOSE 9411
 
-ENTRYPOINT ["java","-Djava.security.egd=file:/dev/./urandom","-jar","/usr/local/bin/zipkin-server.jar"]
+CMD test -n "$STORAGE_TYPE" && source .${STORAGE_TYPE}_profile; java -Djava.security.egd=file:/dev/./urandom -jar zipkin-server.jar
```
adriancole: also notable is that CMD can be overridden w/ docker-compose

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerProperties.java
```diff
@@ -30,10 +30,10 @@ public Store getStore() {
 
   static class Store {
     enum Type {
-      jdbc, inMemory;
+      mysql, mem;
     }
 
-    private Type type = Type.inMemory;
+    private Type type = Type.mem;
```
adriancole: using mem dodges the need to worry about snake case vs camel at least a little longer.

adriancole:
fyi @dsyer @spencergibb

zipkin-java-server/Dockerfile
```diff
@@ -12,16 +12,16 @@
...
 #
 
-FROM openzipkin/zipkin-base:base-1.17.0
+FROM openzipkin/zipkin-base:base-1.17.1
 
 MAINTAINER OpenZipkin "http://zipkin.io/"
 
 VOLUME /tmp
 
-ADD target/zipkin-java-server-0.1.0-SNAPSHOT-exec.jar /usr/local/bin/zipkin-server.jar
```
adriancole: shortened the path as /usr/local/bin doesn't buy us anything except long overrides in docker-compose files :) (implicitly, extending from zipkin-base dumps into /zipkin/)

zipkin-java-server/Dockerfile
```diff
@@ -12,16 +12,16 @@
...
+ADD target/zipkin-java-server-0.1.0-SNAPSHOT-exec.jar zipkin-server.jar
 
-RUN sh -c 'touch /usr/local/bin/zipkin-server.jar'
```
adriancole: sh -c is implicit

zipkin-java-server/Dockerfile
```diff
@@ -12,16 +12,16 @@
...
+RUN touch zipkin-server.jar
 
-EXPOSE 9410 9411
```
adriancole: no longer listen on the collector port (for scribe messages)

adriancole:
@abesto looks like we are basically at the point of drop-in replacement! Yay microservices.

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerProperties.java
```diff
@@ -30,10 +30,10 @@ public Store getStore() {
 
   static class Store {
     enum Type {
-      jdbc, inMemory;
+      mysql, mem;
```
adriancole: mysql since this ties to our other variables, and also allows us to be a drop-in replacement for the query server who also calls this mysql (despite being powered by anorm and jdbc)

abesto:
@adriancole Impressive!

PR #35, opened 2015-10-20
Add support for HTTP POST /api/v1/spans

dsyer:
With no content type (or application/json) we accept JSON, and with
content-type=application/x-thrift we accept Thrift. In both cases
a List<Span> is POSted to /api/v1/spans. The endpoint returns
immediately with a 202 (ACCEPTED) and the spans are written to the
store in an @Async service method.

### DIFF
--- a/zipkin-java-server/src/test/java/io/zipkin/server/ZipkinServerIntegrationTests.java
+++ b/zipkin-java-server/src/test/java/io/zipkin/server/ZipkinServerIntegrationTests.java
@@ -1,0 +1,81 @@
+/**
+ * Copyright 2015 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


zipkin-java-server/src/test/java/io/zipkin/server/ZipkinServerIntegrationTests.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.ConfigurableWebApplicationContext;
+
+import io.zipkin.Annotation;
+import io.zipkin.Codec;
+import io.zipkin.Endpoint;
+import io.zipkin.Span;
+
+/**
+ * @author Dave Syer
```
adriancole: supernit nix

adriancole:
Great job! jealous that this won't likely be as easy upstream :P

dsyer:
(Last commit removes author tag.)

PR #36, opened 2015-10-20
Adds some hardening around accepting content

adriancole:


adriancole:
nothing surprising here, just a fyi @dsyer @spencergibb

zipkin-java-core/src/main/java/io/zipkin/internal/JsonCodec.java
```diff
@@ -47,6 +49,8 @@
  * this should be easy to justify as these objects don't change much at all.
  */
 public final class JsonCodec implements Codec {
+  private static final Logger LOGGER = Logger.getLogger(JsonCodec.class.getName());
```
adriancole: JUL as too early to drag in some version of SLF4J into core (we wouldn't use any of its fanciness here)

PR #37, opened 2015-10-22
Unzips all jar and removes actuator, reducing bootstrap time

adriancole:
With a default docker-machine on a Mid-2014 Macbook Pro, and the default
docker-compose file. I was getting consistent 16s startup, as reported
by Spring Boot's output.

My suspicion was that this was packaging related, as there are many jars
inside the default assembly of Spring Boot. Since Boot does classpath
scanning, it would need to unpack these in order to progress.

@dsyer found that if we do the unpacking at install time, we can reduce
the impact. I verified this, by running the same operation. I'm now
getting consistent 14s ranges on the same docker-machine.

There's other optimizations left. For example, if you remove the
database property and delete the Hikari dep from the pom (so that boot
can't find it), another 2-3 seconds are saved in an otherwise identical
docker-machine.

Regardless, I'm glad to have a couple seconds off without changing
functionality. This change does that.

### DIFF
--- a/zipkin-java-server/Dockerfile
+++ b/zipkin-java-server/Dockerfile
@@ -20,8 +20,8 @@ VOLUME /tmp
 
 ADD target/zipkin-java-server-0.1.0-SNAPSHOT-exec.jar zipkin-server.jar
 
-RUN touch zipkin-server.jar
+RUN unzip zipkin-server.jar
 
 EXPOSE 9411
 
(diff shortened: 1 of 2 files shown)


adriancole:
cc @spencergibb @philwebb

spencergibb:
Interesting, we can turn off classpath scanning, wonder if that would make a difference.

adriancole:
all options welcome! I offer to Q/A any idea :P (just tell me what to c&p
or raise a PR)

PR #39, opened 2015-10-23
Attaches the remote JDBC address as a zipkin "sa" annotation

adriancole:
This is a temporary fix to properly annotate the server our JDBC code
queries.

See https://github.com/openzipkin/brave/issues/104

### DIFF
--- a/zipkin-java-server/src/main/java/io/zipkin/server/brave/JDBCTracerConfiguration.java
+++ b/zipkin-java-server/src/main/java/io/zipkin/server/brave/JDBCTracerConfiguration.java
@@ -14,6 +14,16 @@
 
 package io.zipkin.server.brave;
 
+import com.github.kristofa.brave.Brave;
+import com.mysql.jdbc.Driver;
+import com.twitter.zipkin.gen.AnnotationType;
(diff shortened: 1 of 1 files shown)


adriancole:
This makes sure spans see what the server was, and reinforces latest revisions relating to the dependency graph.

https://github.com/openzipkin/zipkin/pull/790

Happy to pull this out, once we have a cleaner way to do this in Brave.
cc @kristofa @dsyer @spencergibb

zipkin-java-server/src/main/java/io/zipkin/server/brave/JDBCTracerConfiguration.java
```diff
@@ -47,6 +71,11 @@ public void renderEnd(ExecuteContext ctx) {
       this.brave.clientTracer().startNewSpan("query");
       this.brave.clientTracer().setCurrentClientServiceName("zipkin-jdbc");
 
+      // Temporary until https://github.com/openzipkin/brave/issues/104
+      Span span = this.brave.clientSpanThreadBinder().getCurrentClientSpan();
+      synchronized (span) {
```
adriancole: synchronized bc internally this is what brave does

PR #40, opened 2015-10-23
Updates to 1.18.1 plus some tidies

adriancole:
Notably, this updates to properly render the dependency view, based on some things mdsol noticed. This also shows earlier work, where we can display the JDBC destination IP (as it is logged as a "sa" annotation)

screen shot 2015-10-22 at 9 40 57 pm

screen shot 2015-10-22 at 8 50 23 pm

### DIFF
--- a/zipkin-java-interop/pom.xml
+++ b/zipkin-java-interop/pom.xml
@@ -31,9 +31,17 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
-    <zipkin-scala.version>1.18.0</zipkin-scala.version>
+    <zipkin-scala.version>1.18.1</zipkin-scala.version>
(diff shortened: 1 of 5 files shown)


adriancole:
cc @adrianco @jcarres-mdsol (note this is the java project, not the scala one)

zipkin-java-interop/pom.xml
```diff
@@ -64,7 +72,8 @@
       <groupId>io.zipkin</groupId>
       <artifactId>zipkin-common</artifactId>
       <classifier>test</classifier>
-      <version>${zipkin-scala.version}</version>
+      <!-- TODO revert to ${zipkin-scala.version} on 1.18.2 -->
+      <version>1.18.2-SNAPSHOT</version>
```
adriancole: some upstream tests needed adjustment

Issue #38, opened 2015-10-22
Implement MYSQL_MAX_CONNECTIONS


adriancole:
Not sure how to plug this in, but we should be able to change HikariCP's max connections value with MYSQL_MAX_CONNECTIONS

https://github.com/openzipkin/zipkin/pull/789

dsyer:
Some the configuration options are listed here: http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#common-application-properties. Anything that is a Java Bean property of the `DataSource` can be set using `spring.datasource.*`. Probably `spring.datasource.maxActive=...` is what you want for this.

adriancole:
I thought I tried that.. maybe it was the capital that I missed last night.
Thx!

PR #41, opened 2015-10-23
Switch to Tomcat data source in place of Hikari

dsyer:
This change does not compromise functionality but it speeds up
start up (shaves approx. 1s off 4s startup time on my laptop).

### DIFF
--- a/zipkin-java-server/pom.xml
+++ b/zipkin-java-server/pom.xml
@@ -74,8 +74,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.springframework</groupId>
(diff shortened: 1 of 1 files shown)


adriancole:
Thanks for the workaround. Seconds in laptop get multiplied out in virtualized environments, so this does matter.

Hopefully, we can figure out what causes the performance issues as I've no doubt tomcat is larger, more complex configuration and performs worse than Hikari. Regardless, that's not today's issue as we are using defaults and this is still a pre 1.0.

By the way, I logged this yesterday. Any ideas on how to get this mapped to Tomcat's config? https://github.com/openzipkin/zipkin-java/issues/38

PR #42, opened 2015-10-27
Updates to upstream 1.19.1, eliminating the collector dependency

adriancole:
Starting with zipkin 1.19.1, the UI process (zipkin-web) no longer
depends on scribe for self-tracing. This cuts the minimum deployment to
2 processes, which happen to be the smallest.

The net effect is a simpler starter architecture, which takes less time
to understand, less MB to download, and less memory to run.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,20 +13,9 @@
 #
 
 mysql:
-  image: openzipkin/zipkin-mysql:1.18.0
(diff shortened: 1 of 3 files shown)


adriancole:
@dsyer @spencergibb fyi!

adriancole:
tested in a local and aws docker-machine

Issue #7, opened 2015-09-08
Publishing


adriancole:
We should publish snapshots and releases, snapshots first

cc @kristofa @abesto @jamescway

kristofa:
@adriancole @abesto @jamescway 
Because I have experience with publishing to Sonatype I would follow [this approach](http://central.sonatype.org/pages/ossrh-guide.html). 
Or do you prefer using bintray like we did for the other artefacts?  Do we sync to Maven Central already? Do we already have a Sonatype account?

adriancole:
@kristofa I'm fine using sonatype directly, at least as a start. You can
ask for access here (I've been using my own creds for zipkin pushes)
https://issues.sonatype.org/browse/OSSRH-16669

kristofa:
@adriancole Good! The issue you mention is closed. If I add a comment with my user name will they still process it?

adriancole:
They have been, so give it a try! I can also share my creds with you and/or
otherwise loan them to the process.

kristofa:
@adriancole I'll update the issue and give it a try. No need to share creds. I can prepare a pr with `pom.xml` changes and you can test it out if I don't get access soon enough :)

adriancole:
Awesome

adriancole:
Maybe next step is to setup travis to auto-publish snapshots? Ex. https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/.travis.yml

jamescway:
@adriancole @rtyler can you contact me if you have the openzipkin credentials?  Presumably there is also an email address associated with the account.

adriancole:
I don't have an OpenZipkin account. I have only used my sonatype account in
the past.

adriancole:
looks like @jamescway got round 1 complete! we have snapshots.

adriancole:
ok now snapshots are going to jfrog, like our other ones: http://oss.jfrog.org/artifactory/oss-snapshot-local/io/zipkin/zipkin-java/0.1.0-SNAPSHOT/

adriancole:
rationale for switching from sonatype to jfrog snapshots is that we don't want users to have confusion by requiring them to go to different artifactory instances depending on the primary language of the project.

adriancole:
0.1.0 is out on maven central, so this is quite a milestone!

The following will now publish to bintray: `./mvnw  -s ./.settings.xml release:prepare`
* Note: bintray does the signing, so you don't need to setup GPG.

[...]

PR #44, opened 2015-11-15
Switches default JDBC driver from MySQL to MariaDB

adriancole:
MySQL connector is GPL, even if it has an OSS exception.
https://www.mysql.com/about/legal/licensing/foss-exception/

MariaDB has a friendlier license, LGPL, which is less scary in audits.


adriancole:
cc @mstine @abesto @dsyer @spencergibb

zipkin-java-server/src/main/resources/zipkin-server.yml
```diff
@@ -11,7 +11,7 @@ server:
   port: ${QUERY_PORT:9411}
 spring:
   datasource:
-    driver-class-name: com.mysql.jdbc.Driver
+    driver-class-name: org.mariadb.jdbc.Driver
     url: jdbc:mysql://${mysql.host}:${mysql.port}/${mysql.db}?autoReconnect=true
```
adriancole: note: mariadb supports the same URI as the mysql driver

abesto:
Are there any tech trade-offs we need to be aware of? (Disclaimer: I have zero experience with MariaDB)

adriancole:
This is a different driver than the MySQL connector, and while we don't use extended flags there would be differences in how things like reconnect work. In both cases, these are drivers backed by a commercial company.

In the `zipkin-java-jdbc` project, unlike `zipkin-anormdb`, there's no tight coupling to a specific driver, so someone could swap it out to Oracle's driver. There's only one impact: self-tracing of JDBC is currently conditional on the driver type: We could probably move to a different condition, but regardless self-tracing of zipkin is more a showcase feature.

abesto:
Alright, then this is a :+1: from me.

Let's make sure we get semver right when releasing; I guess minor would be the right call here.

adriancole:
haha... this has never had a release! but yeah, sure thing on upstream

abesto:
Oh right, we're on `zipkin-java`. Oops :)

adriancole:
ex. #7 is incomplete.. definitely want to release this week

PR #46, opened 2015-11-17
Adds bootstrap and codec to self-tracing

adriancole:
The scala project logs a local trace of bootstrap, to serve as an
example of how to do that. This adds a local trace, including the same
start signal (init), but spring-boot specific annotations for lifecycle
events.

This also adds a SpanStore tracer (except for writes), to show an
example of how to trace a local component

Here's an example bootstrap trace from docker on my macbook pro:

screen shot 2015-11-17 at 8 41 06 am

And here's an example of how much time zipkin spends actually retrieving spans (vs http/api overhead):

screen shot 2015-11-23 at 7 12 00 am

This change is awaiting a release of brave 3.2

### DIFF
--- a/zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerConfiguration.java
+++ b/zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerConfiguration.java
@@ -13,10 +13,12 @@
  */
 package io.zipkin.server;
 
+import com.github.kristofa.brave.Brave;
 import io.zipkin.Codec;
 import io.zipkin.SpanStore;
 import io.zipkin.jdbc.JDBCSpanStore;
 import io.zipkin.server.ZipkinServerProperties.Store.Type;
+import io.zipkin.server.brave.TraceWritesSpanStore;
 import javax.sql.DataSource;
 import org.jooq.ExecuteListenerProvider;
 import org.jooq.conf.Settings;
@@ -43,6 +45,9 @@ public class ZipkinServerConfiguration {
   @Qualifier("jdbcTraceListenerProvider")
   ExecuteListenerProvider listener;
 
+  @Autowired(required = false)
+  Brave brave;
+
   @Bean
   @ConditionalOnMissingBean(Codec.Factory.class)
   Codec.Factory codecFactory() {
@@ -51,10 +56,12 @@ public class ZipkinServerConfiguration {
 
   @Bean
   SpanStore spanStore() {
+    SpanStore result;
     if (this.datasource != null && this.server.getStore().getType() == Type.mysql) {
-      return new JDBCSpanStore(this.datasource, new Settings().withRenderSchema(false), this.listener);
+      result = new JDBCSpanStore(this.datasource, new Settings().withRenderSchema(false), this.listener);
     } else {
-      return new InMemorySpanStore();
+      result = new InMemorySpanStore();
     }
+    return brave != null ? new TraceWritesSpanStore(brave, result) : result;
   }
 }
--- a/zipkin-java-server/src/main/java/io/zipkin/server/brave/TraceWritesSpanStore.java
+++ b/zipkin-java-server/src/main/java/io/zipkin/server/brave/TraceWritesSpanStore.java
@@ -1,0 +1,98 @@
+/**
+ * Copyright 2015 The OpenZipkin Authors
+ *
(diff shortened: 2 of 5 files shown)


adriancole:
@dsyer @spencergibb @kristofa @philwebb any notes on the code before I backfill tests?

zipkin-java-server/src/main/java/io/zipkin/server/brave/JDBCTracerConfiguration.java
```diff
@@ -60,7 +60,7 @@ Endpoint jdbcServerAddr(@Value("${spring.datasource.url}") String jdbcUrl) throw
   public void renderEnd(ExecuteContext ctx) {
     if (ctx.type() == ExecuteType.READ) { // Don't log writes (so as to not loop on collector)
       this.brave.clientTracer().startNewSpan("query");
-      this.brave.clientTracer().setCurrentClientServiceName("zipkin-jdbc");
+      this.brave.clientTracer().setCurrentClientServiceName("zipkin-query");
```
adriancole: unrelated bug

adriancole:
I'm going to add codec tracing to this next

adriancole:
Updated with a local trace at the end of the tree. @jcarres-mdsol ^^ is an example of where something <1ms can show up.

adriancole:
as these are internal classes, I'll go ahead and shine this up with tests for merge. We can change this when we figure out a different impl and/or when we add local trace support to brave.

adriancole:
This is dependent on https://github.com/openzipkin/brave/pull/112 cc @kristofa

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerConfiguration.java
```diff
@@ -51,10 +56,12 @@
...
   SpanStore spanStore() {
+    SpanStore result;
     if (this.datasource != null && this.server.getStore().getType() == Type.mysql) {
-      return new JDBCSpanStore(this.datasource, new Settings().withRenderSchema(false), this.listener);
+      result = new JDBCSpanStore(this.datasource, new Settings().withRenderSchema(false), this.listener);
     } else {
-      return new InMemorySpanStore();
+      result = new InMemorySpanStore();
     }
+    return brave != null ? new TraceWritesSpanStore(brave, result) : result;
```
adriancole: @dsyer @spencergibb I couldn't figure out how to make BraveConfiguration override or decorate a binding. If there was, I could move this to the brave package, as doing it here feels like cheating.

zipkin-java-server/src/main/java/io/zipkin/server/brave/TraceWritesSpanStore.java
```diff
@@ -0,0 +1,98 @@
...
+    } finally {
+      tracer.finishSpan();
+    }
+  }
+
+  @Override
+  public List<List<Span>> getTracesByIds(List<Long> traceIds) {
+    tracer.startSpan(component, "get-traces-by-ids");
+    try {
+      return delegate.getTracesByIds(traceIds);
```
adriancole: It may appear that we should extract a method that takes a Func1 method reference and its arg, but we'd also have to do that for Func0 (Supplier). I counted the lines and there's no advantage to doing that, even if it looks neat. The advantage of not trying to extract methods for these is that we can add binary annotations as needed without even more indirection. Look ^^ on getTraces for example.

zipkin-java-server/src/main/java/io/zipkin/server/brave/TraceWritesSpanStore.java
```diff
@@ -0,0 +1,98 @@
...
+    } finally {
+      tracer.finishSpan();
+    }
+  }
+
+  @Override
+  public List<DependencyLink> getDependencies(long endTs, @Nullable Long lookback) {
+    tracer.startSpan(component, "get-dependencies");
+    try {
+      return delegate.getDependencies(endTs, lookback);
```
adriancole: ex maybe we should log a binary annotation for the period of lookback, or the inputs, as this could be really long (and notable in a trace).

adriancole:
This PR will likely wait longer than my usual max 3 days until merge or close as I'm moving and we need brave 3.2. Any feedback in the mean time is love, but be patient w/response!

adriancole:
found a glitch needed before we can merge this: https://github.com/openzipkin/brave/pull/114

PR #47, opened 2015-11-17
Moves codec ops to an interface in preparation of tracing

adriancole:
Codec had a larger interface than needed, and instances were used
directly. This moves to an interface for reasons including preparation
of codec tracing via dependency injection.


spencergibb:
When removed lines are greater than modified :clap:.

PR #50, opened 2015-12-11
Make brave optional

dsyer:
Brave is an extra dependency, and some people might prefer not
to use it. The change to make it optional is fairly trivial
so it seems worth it.


spencergibb:
:+1:

adriancole:
Thanks! I was wondering how to do this

PR #51, opened 2015-12-21
Adjusts environment variables to match zipkin-scala and adds docs

adriancole:
This adds docs regarding which environment variables are supported. It
also corrects `MYSQL_PORT` which is actually `MYSQL_TCP_PORT`.


adriancole:
tested MYSQL_USE_SSL against AWS RDS

PR #54, opened 2016-01-03
Automagically deal with copyright year updates

adriancole:
Per CONTRIBUTING.md, license headers are updatable via `./mvnw com.mycila:license-maven-plugin:format`. The impact of this config is that the first update to a file per year needs to change the range. Ex. `2015` to `2015-2016`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015 The OpenZipkin Authors
+    Copyright 2015-2016 The OpenZipkin Authors
 
     Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
     in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015 The OpenZipkin Authors
+    Copyright 2015-2016 The OpenZipkin Authors
```
adriancole: ex. I touched this file

src/etc/header.txt
```diff
@@ -1,4 +1,4 @@
-Copyright 2015 The OpenZipkin Authors
+Copyright ${license.git.copyrightYears} The OpenZipkin Authors
```
adriancole: The license verification plugin will not expect all files to have the same copyright. Instead, it looks at git history.

adriancole:
cc @abesto nifty, huh?

adriancole:
merged this, as otherwise adding new code needs 2015, or we need to switch all files to 2016!

abesto:
Awesome. I'm impressed that someone(s) actually took the time to do this right. (Then again, Java is used for stuff where this matters, so I guess I shouldn't be surprised)

PR #57, opened 2016-01-04
Honor COLLECTOR_SAMPLE_RATE with TraceIdSampler

### LINKED ISSUE #52: Add server-side sampler
In order to be compatible with zipkin-scala's collector, we need to implement sampling, particularly the following:

`COLLECTOR_SAMPLE_RATE`: Percentage of traces to retain, defaults to always sample (1.0).

This is a rate that applies when Span.debug is not true, and treats the sample rate as a percentage. Ex. .1 means retain 10%

See https://github.com/openzipkin/zipkin/blob/master/zipkin-collector-service/src/main/scala/com/twitter/zipkin/collector/sampler/AdjustableGlobalSampler.scala

adriancole:
This introduces a port of the zipkin-scala sampler as TraceIdSampler.
This is used to implement the `COLLECTOR_SAMPLE_RATE` env property.

In order to make filtering easy and portable with language level 7, this
changes `SpanStore.accept` to take an iterator as opposed to a list.

Fixes #52

### DIFF
--- a/zipkin-java-core/pom.xml
+++ b/zipkin-java-core/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015 The OpenZipkin Authors
+    Copyright 2015-2016 The OpenZipkin Authors
 
     Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
     in compliance with the License. You may obtain a copy of the License at
@@ -28,8 +28,6 @@
   <description>Zipkin Java Core</description>
 
   <properties>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
     <main.basedir>${project.basedir}/..</main.basedir>
     <animal-sniffer-maven-plugin.version>1.14</animal-sniffer-maven-plugin.version>
   </properties>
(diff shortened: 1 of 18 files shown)


zipkin-java-core/pom.xml
```diff
@@ -47,6 +45,28 @@
 
   <build>
     <plugins>
+      <plugin>
+        <inherited>true</inherited>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
```
adriancole: this allows lambdas in test tree

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinQueryApiV1.java
```diff
@@ -44,7 +45,10 @@
 public class ZipkinQueryApiV1 {
 
   private static final String APPLICATION_THRIFT = "application/x-thrift";
-  private static final String DEFAULT_LOOKBACK = "86400000"; // 7 days in millis
+
+  @Autowired
+  @Value("${zipkin.query.lookback}")
```
adriancole: unrelated, but fixed anyway (also tested manually)

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinServerProperties.java
```diff
@@ -43,16 +38,4 @@ public void setType(Type type) {
       this.type = type;
     }
   }
-
-  static class Query {
```
adriancole: dead

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinSpanWriter.java
```diff
@@ -15,14 +15,29 @@
...
+  @Autowired
+  TraceIdSampler sampler;
+
+  /**
+   * Asynchronously writes spans to storage, subject to sampling policy.
+   */
   @Async
   public void write(SpanStore spanStore, List<Span> spans) {
-    spanStore.accept(spans);
+    Iterator<Span> sampled = spans.stream()
```
adriancole: this is the actual collector sampling policy as implemented in zipkin-scala

zipkin-java-server/src/main/resources/zipkin-server.yml
```diff
@@ -7,6 +7,9 @@ mysql:
   max-active: ${MYSQL_MAX_CONNECTIONS:10}
   use-ssl: ${MYSQL_USE_SSL:false}
 zipkin:
+  collector:
+    # percentage to traces to retain
+    sample-rate: ${COLLECTOR_SAMPLE_RATE:1.0}
```
adriancole: this is the env variable used in zipkin-scala and docker

adriancole:
@dsyer @spencergibb @marcingrzejszczak better late than never!! I'll release this as v0.2 tomorrow GMT+8

PR #59, opened 2016-01-07
Removes Closeable contract from SpanStore

adriancole:
Closeable isn't the only way to close things, and in reality we weren't
doing anything with it. This punts the decision on how to close
resources to implementing types.

See https://github.com/openzipkin/zipkin/issues/584


adriancole:
cc @kristofa

kristofa:
:+1:

PR #60, opened 2016-01-08
Fixes imports per style and removes unnecessary "this" qualification

adriancole:
[Spring Framework style guide](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style) requires folks to always qualify fields
with `this.`. This project isn't a part of the spring framework, nor
derives it style from that.

Square and Google style (from which it derives) does not require such
qualification. I've contributed to projects from many orgs, and Spring
is the only place I've seen this. I don't think it is a stretch to call
this unconventional.

Unconventional is fine, when it improves the experience of development.
Writing much of the code in this codebase, I've found the unnecessary
qualification to cause quite the opposite. It is a cognitive break,
where I wonder what is special which requires this qualification? Other
times, it intrudes on readability, for example chaining with parameters
ends up with `this.` boilerplated throughout the calls. In other words,
the fluency is reduced. Minor, but this can make long chains worse, or
cause wrapping.

While I feel it wouldn't be in my place to ask for Spring to change its
style, including use of tabs and `this.` qualifications, I do think it
is reasonable to not bring this (no pun intended) here.

The result of reverting this change, if convention isn't valuable, will
at least be a more joyful experience for the primary contributor.

### DIFF
--- a/zipkin-java-core/src/main/java/io/zipkin/Annotation.java
+++ b/zipkin-java-core/src/main/java/io/zipkin/Annotation.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 The OpenZipkin Authors
+ * Copyright 2015-2016 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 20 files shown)


adriancole:
Raising this as a pull request as an FYI. Apologies to those who use things that require `this.` speckling, even if they don't personally like it.

cc'ing those who contribute or who've expressed interest in doing so.
@kristofa @dsyer @marcingrzejszczak @michaelsembwever @spencergibb

PR #62, opened 2016-01-09
Renames TraceIdSampler to be similar to opentracing/api-python

adriancole:
The OpenTracing python api has a zipkin implementation with an almost
identical sampler as `TraceIdSampler`. This renames things so that it
isn't artificially different.

See https://github.com/opentracing/api-python/blob/master/example/zipkin_like/sampler.py


adriancole:
@kristofa heads up.. I'll be making an identical change in brave.
@marcingrzejszczak also heads up. I suspect sleuth will want to name things similarly

@yurishkuro fyi

zipkin-java-core/src/main/java/io/zipkin/TraceIdSampler.java
```diff
@@ -1,106 +0,0 @@
...
- * is made before any work is measured, or annotations are added. As such, the input parameter to
- * zipkin v1 samplers is the trace id (64-bit random number).
- *
- * <p>The implementation is based on zipkin-scala's AdjustableGlobalSampler. It is percentage based,
- * and its accuracy is tied to distribution of traceIds across 64bits. For example, tests have shown
- * an error rate of 3% when traceIds are random. It is idempotent, ie a consistent decision for a
- * given trace id.
- */
-// abstract for factory-method support on Java language level 7
-public abstract class TraceIdSampler {
```
adriancole: github didn't like all the renames!! impl is the same

PR #65, opened 2016-01-13
Add default values to @Value annotations where possible

dsyer:
Without these you can't use zipkin-java-server as a library without
special knowledge of the YAML configuration. With them it is possible
to start a server without any configuration at all.

### DIFF
--- a/zipkin-java-server/src/main/java/io/zipkin/server/ZipkinQueryApiV1.java
+++ b/zipkin-java-server/src/main/java/io/zipkin/server/ZipkinQueryApiV1.java
@@ -13,13 +13,12 @@
  */
 package io.zipkin.server;
 
-import io.zipkin.Codec;
-import io.zipkin.QueryRequest;
-import io.zipkin.Span;
-import io.zipkin.SpanStore;
+import static io.zipkin.internal.Util.checkNotNull;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
 import java.util.Collections;
 import java.util.List;
-import okio.Buffer;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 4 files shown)


zipkin-java-server/src/main/java/io/zipkin/server/ZipkinQueryApiV1.java
```diff
@@ -47,7 +49,7 @@
   private static final String APPLICATION_THRIFT = "application/x-thrift";
 
   @Autowired
-  @Value("${zipkin.query.lookback}")
+  @Value("${zipkin.query.lookback:86400000}")
   int defaultLookback = 86400000; // 7 days in millis
```
marcingrzejszczak: I guess this value is not needed any more if you provide it via `@Value` ?

dsyer: Technically true, but only in a Spring context. There might be unit tests that need a value (there aren't, but I was leaving it initialized to be defensive).

marcingrzejszczak: +1

zipkin-java-server/src/main/java/io/zipkin/server/ZipkinQueryApiV1.java
```diff
@@ -111,7 +113,7 @@ public void uploadSpansThrift(@RequestBody byte[] body) {
         .minDuration(minDuration)
         .maxDuration(maxDuration)
         .endTs(endTs)
-        .lookback(lookback != null ? lookback : defaultLookback)
+        .lookback(lookback != null ? lookback : this.defaultLookback)
```
adriancole: argh.. this. is back

adriancole:
I'm going to pick this in, scrubbing the `this.` and import changes. Thanks for the info on how to get this to work!

adriancole:
Thanks for fixing this. releasing now as 0.4.1

PR #71, opened 2016-01-19
Removes null coersion in Codec in favor of IllegalArgumentException

### LINKED ISSUE #66: Improve Codec.readSpans debuggability
Especially when spans are encoded into json, we can report better feedback than just a 400 code. Particularly as new instrumentation is written, there will be cases where spans are malformed in some way or another. We want people to solve things quickly.

This means re-jigging the contract of codec to throw instead of return null. This will hurt code readability when used in lambdas, as lambdas can't throw. However, it is a fair tradeoff.

I will try my best to give context, eventhough this is hard in thrift. One way is to keep a position in the data, and also summary of which trace/spanids were successfully decoded (if any).

When spans are completely malformed (ex. gzipped or sent in wrong encoding), we may not be able to give a useful message back. However, in cases where fields are missing, etc. we could do better.

This is to help folks like @marcingrzejszczak who have had an unnecessarily difficult time troubleshooting.

adriancole:
When a codec is reading, it is from a byte array, so the only possiblity
of problems is malformed input. This uses IllegalArgumentException to
indicate something the user can fix.

Since all domain objects are locally defined in core, and validated at
construction time, any errors writing to a byte array are bugs. This
uses AssertionError to indicate a bug.

Before, we weren't sending an error message to the user. This sends a
single-line description back in the 400 response. It also moves logging
to the call-site (`ZipkinQueryApiV1`).

Example curl request with bad content:
```bash
$ curl -X POST http://localhost:9411/api/v1/spans -H 'Content-Type: application/x-thrift' -d'foobar'
Malformed reading List<Span> from TBinary: Zm9vYmFy
```

Fixes #66

### DIFF
--- a/zipkin-java-core/src/test/java/io/zipkin/CodecTest.java
+++ b/zipkin-java-core/src/test/java/io/zipkin/CodecTest.java
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 The OpenZipkin Authors
+ * Copyright 2015-2016 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -15,12 +15,16 @@ package io.zipkin;
 
 import java.io.IOException;
 import java.util.List;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public abstract class CodecTest {
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
 
   protected abstract Codec codec();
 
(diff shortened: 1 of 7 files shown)


zipkin-java-core/src/test/java/io/zipkin/CodecTest.java
```diff
@@ -80,8 +63,53 @@ public void dependencyLinksRoundTrip() throws IOException {
   }
 
   @Test
-  public void dependencyLinksDecodeToNullOnEmpty() throws IOException {
-    assertThat(codec().readDependencyLinks(new byte[0]))
-        .isNull();
+  public void decentErrorMessageOnEmptyInput_span() throws IOException {
+    thrown.expect(IllegalArgumentException.class);
+    thrown.expectMessage("Empty input reading Span");
```
adriancole: chose to repeat 3 times as both the message and the method references vary. we could get clever, but decided to go less fancy.

marcingrzejszczak:
Sweet

PR #72, opened 2016-01-20
Reorganizes maven artifacts under io.zipkin.java, repackages as "zipkin"

adriancole:
This reorganizes the maven project so that it is easier to add new
features, such as a Cassandra SpanStore, without bumping into existing
artifacts from the scala project. It does this by placing all artifacts
under a package corresponding to the repository: io.zipkin.java.

This also decouples the java package hierarchy from the domain name. By
using the package "zipkin" as opposed to "io.zipkin", when we change the
model in an incompatible way, say version 2, we can rename it to 
"zipkin2", where "io.zipkin2" makes less sense. This is the convention
used in OkHttp and Retrofit. It is also similar to HTrace's approach,
which also makes the major version a component of the package name.

This also introduces some other naming conventions taken from Retrofit.

For example, root directories that are not prefixed with zipkin are not
published. For example, interop and benchmark modules are not published,
while zipkin-spanstores are.

As the repo is now namespaced properly via the project id, artifacts can
have simpler names, for example "zipkin" instead of "zipkin-java-core".
Moreover, we now don't mix artifacts in the same namespace as the scala
ones. This would lead to less confusion about a spanstore-cassandra in
the future.

See http://jakewharton.com/java-interoperability-policy-for-major-version-updates/


adriancole:
christmas build is due to https://github.com/openzipkin/zipkin/issues/878

adriancole:
I verified that only important artifacts are now installed

```bash
$ find /Users/acole/.m2/repository/io/zipkin/java -name \*jar
/Users/acole/.m2/repository/io/zipkin/java/spanstore-jdbc/0.4.3-SNAPSHOT/spanstore-jdbc-0.4.3-SNAPSHOT.jar
/Users/acole/.m2/repository/io/zipkin/java/zipkin/0.4.3-SNAPSHOT/zipkin-0.4.3-SNAPSHOT.jar
/Users/acole/.m2/repository/io/zipkin/java/zipkin-server/0.4.3-SNAPSHOT/zipkin-server-0.4.3-SNAPSHOT-exec.jar
/Users/acole/.m2/repository/io/zipkin/java/zipkin-server/0.4.3-SNAPSHOT/zipkin-server-0.4.3-SNAPSHOT.jar
```

adriancole:
Also ran benchmarks to make sure the instructions still work.

```bash
$ java -jar benchmarks/target/benchmarks.jar
# JMH 1.11.3 (released 6 days ago)
# VM version: JDK 1.8.0_51, VM 25.51-b03
# VM invoker: /Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk/Contents/Home/jre/bin/java
--snip--
```

adriancole:
Also checked spring boot, both in-memory `./mvnw -pl zipkin-server spring-boot:run` and mysql `MYSQL_USER=root STORAGE_TYPE=mysql ./mvnw -pl zipkin-server spring-boot:run`

adriancole:
Most of these conventions were taken from okhttp and retrofit, which recently re-orged things and summarized rationale here: http://jakewharton.com/java-interoperability-policy-for-major-version-updates/

adriancole:
released as 0.4.4

michaelsembwever:
+1

PR #73, opened 2016-02-03
Re-organizes Node so that it can be used on objects besides Span

adriancole:
We need to create the trace tree in dependency store code. Particularly,
we don't want to retrieve the entire span just to count calls. This
refactors the Node object so that it can be used when all span details
aren't present.


adriancole:
@t-yuki @marcingrzejszczak fyi, this is in preparation for fixing the dependency graph

PR #74, opened 2016-02-05
Updates to latest zipkin-scala and adds in-mem dependency aggregation

adriancole:
Previously, the in-memory provider did not support dependency
aggregation. This meant users couldn't play with zipkin unless they ran
MySQL. This adds aggregation support through a new type:
`DependencyLinksBuilder`.

This new type also corects the dependency graph, particularly around
uninstrumented services or paths with local spans intermediating calls.

### DIFF
--- a/interop/pom.xml
+++ b/interop/pom.xml
@@ -29,7 +29,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
-    <zipkin-scala.version>1.30.2</zipkin-scala.version>
+    <zipkin-scala.version>1.32.3</zipkin-scala.version>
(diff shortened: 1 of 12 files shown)


adriancole:
cc @jcarres-mdsol @t-yuki @marcingrzejszczak after this change, local spans work on the dependency graph (here in the java project), both in-memory and also with mysql

adriancole:
rewrote the dependency link SQL to use a single select instead of two. This uses a left join to avoid returning strings we aren't interested in.

I plan to refactor the java code a little more before merging.

adriancole:
code is in the basic shape that I want. there's a little optimizing left and documentation fixes. will probably wrap up tomorrow.

PR #76, opened 2016-02-11
Fix fragile test

dsyer:
This test was failing for me and it looks a bit fragile because it
is sensitive to whitespace at the end of a String. This change
makes it more robust.


adriancole:
Thanks. I'll polish this on the way in as only one line of change (plus an import) seem to be the focus.

PR #77, opened 2016-02-11
Add a minimal sample as an integration test

dsyer:
Since some of the dependencies in the main server module are optional it
makes sense to have a test that user can build a server without them. This
change adds such a module, so tests will verify that the dependencies are
indeed optional.


adriancole:
kindof prefer a different approach as I don't want to advertise one-off servers. mind if I change the packaging such that it is more explicitly a test vs a recommendation?

adriancole:
for example, I don't want to publish this as a dist, but certainly agree since we use optional deps we ought to test that they are optional.

dsyer:
Other than changing the name, I'm not sure what more you could do to make it explicitly a test. What did you have in mind?

adriancole:
one option is to move the tree under zipkin-server/src/it and use invoker
to run it. This is explicitly a test and I think it won't require much to
exclude it from publication etc.

ex. rename it to zipkin-server/src/it/deps-are-optional

This is how dagger does it, for example.
https://github.com/google/dagger/tree/master/compiler/src/it

dsyer:
I pushed it down into a src/it test (but only activated with a profile for now, so it doesn't slow down the "normal" build).

adriancole:
thanks.. I'm going to pare this down a bit as we only really need to exercise the server (don't need a README that includes how to run as we aren't advertising, etc). Also, I suspect we want this profile to be run in travis, so I'll add that.

PR #81, opened 2016-02-22
Adds ZipkinRule

### LINKED ISSUE #64: MockZipkinServer
I want to write something to help ease testing. Basically a dependency-light hard-wired junit rule, which avails a zipkin query endpoint and a reference to the in-memory span store.

This would be used the same way people use CassandraUnit or KafkaRule, or.. MockWebServer

We can build this directly on `MockWebServer`, as it is dependency light, does a good job at things like picking ports, and its dispatcher is pluggable (I've done this in the past). Configuration would be fixed, ex it can only use the InMemorySpanStore, so it won't need to bring in dependencies like spring (which could lead to conflicts in the test classpath).

adriancole:
This adds ZipkinRule, a JUnit rule to spin-up a Zipkin server during tests. For example, you can write micro-integration tests like so:

```java
@Rule
public ZipkinRule zipkin = new ZipkinRule();

// Pretend we have a traced system under test
TracedService service = new TracedService(zipkin.httpUrl(), ReportingMode.FLUSH_EVERY);

@Test
public void skipsReportingWhenNotSampled() throws IOException {
  zipkin.storeSpans(asList(rootSpan));

  // send a request to the instrumented server, telling it not to sample.
  client.addHeader("X-B3-TraceId", rootSpan.traceId)
        .addHeader("X-B3-SpanId", rootSpan.id)
        .addHeader("X-B3-Sampled", 0).get(service.httpUrl());

  // check that zipkin didn't receive any new data in that trace
  assertThat(zipkin.getTraces()).containsOnly(asList(rootSpan));
}
```

Fixes #64

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,6 +24,7 @@
 
   <modules>
     <module>zipkin</module>
+    <module>zipkin-junit</module>
     <module>benchmarks</module>
     <module>interop</module>
     <module>zipkin-spanstores</module>
@@ -122,6 +123,12 @@
         <groupId>${project.groupId}</groupId>
         <artifactId>zipkin</artifactId>
         <type>test-jar</type>
+        <version>${project.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>zipkin-junit</artifactId>
         <version>${project.version}</version>
       </dependency>
 
--- a/zipkin-junit/README.md
+++ b/zipkin-junit/README.md
@@ -1,0 +1,26 @@
+# zipkin-junit
+
+This contains `ZipkinRule`, a JUnit rule to spin-up a Zipkin server during tests.
+
+For example, you can write micro-integration tests like so:
+
+```java
+@Rule
+public ZipkinRule zipkin = new ZipkinRule();
+
+// Pretend we have a traced system under test
+TracedService service = new TracedService(zipkin.httpUrl(), ReportingMode.FLUSH_EVERY);
+
+@Test
+public void skipsReportingWhenNotSampled() throws IOException {
+  zipkin.storeSpans(asList(rootSpan));
+
+  // send a request to the instrumented server, telling it not to sample.
+  client.addHeader("X-B3-TraceId", rootSpan.traceId)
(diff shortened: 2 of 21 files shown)


adriancole:
This MockZipkinServer can be used to test zipkin instrumentation projects.

This doesn't have deps like spring, which means you can test spring instrumentation without interference.
It also doesn't do anything asynchronously, which means you don't need retry loops in tests.

right now, this implementation is bare-bones (only implements the http api and tests it). I suspect in some future, we could make fancy assertions on top (ex. id insensitive assertions, or tree-based ones).

I made the forwarding OkHttpSpanStore to test the mock server, and those test pass 100%.

Anyway, let me know what you think.. If the current shape is ok, I'll tidy up with README etc.

cc @kristofa @marcingrzejszczak  @dsyer @CodingFabian @michaelsembwever

zipkin-spanstores/okhttp/src/main/java/zipkin/okhttp3/OkHttpSpanStore.java
```diff
@@ -0,0 +1,141 @@
...
+import zipkin.QueryRequest;
+import zipkin.Span;
+import zipkin.SpanStore;
+import zipkin.internal.JsonCodec;
+import zipkin.internal.Nullable;
+
+/**
+ * Implements the span store interface by forwarding requests over http.
+ */
+public class OkHttpSpanStore implements SpanStore {
```
adriancole: ps I'm considering flattening this class into the mockzipkinserver module's test tree, as opposed to a separate spanstore-okhttp module. If it is later asked about directly, it could be bumped out like this then. Unless someone screams.. I'll probably remove the spanstore-okhttp module tomorrow.

marcingrzejszczak:
Looks really nice! And will be definitely very useful in integration tests.

kristofa:
A http MockZipkinServer is definitely useful!  I don't know OkHttp but I'll soon look into it (I might use the mock server for testing failure cases like time outs)

adriancole:
renamed to ZipkinRule to be more like other rules like kafka and cassandra. cc @clehene

Polishing tests and will merge today

PR #85, opened 2016-02-25
Removes internal use of libthrift

adriancole:
This implements TBinaryProtocol directly in the ThriftCodec. By doing
the code, dependencies, and exceptions are simpler (as we are decoding
via a buffer not a network). This uses Okio Buffer, which also
simplified some code.

The payment was having to implement a couple utilities from libthrift,
namely the skip logic. Overall, the code is smaller and the binary
shrunk from 275k to 225k.

### DIFF
--- a/interop/pom.xml
+++ b/interop/pom.xml
@@ -29,7 +29,9 @@
 
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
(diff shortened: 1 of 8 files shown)


adriancole:
cc @codingfabian @kristofa

CodingFabian:
pretty nice. thank you.

PR #87, opened 2016-02-25
Compresses GET responses

adriancole:
We were missing a parameter that controls GET compression. With this
change, json responses will be compressed.

Note: In spring-boot, only responses over a threshold are gzipped:
`server.compression.min-response-size` defaults to 2KiB.


adriancole:
@dsyer @marcingrzejszczak fyi! I'll merge this after writing a unit test, but this does pass manually

adriancole:
I added tests for the zipkin rule, but `ZipkinServerIntegrationTests` doesn't seem to honor the compression setting. I've tested it manually via below:

```
$ curl --compress -sv localhost:9411/api/v1/trace/7f745fff06fc7860 -H 'Accept-Encoding: gzip'
```

PR #88, opened 2016-02-29
Merging of already existing spans

marcingrzejszczak:
* the in memory span is merging existing spans
* the mysql one requires spans to arrive in a very precise order which in case of amqp for instance is not guaranteed

TODO:
* retention of the cache

Relates to: https://github.com/spring-cloud/spring-cloud-sleuth/issues/188


marcingrzejszczak:
this approach is wrong unfortunately. We would have to have a distributed cache...

adriancole:
agreed, this isn't the best way as you'd need to consistently route to the same collection instance per trace

PR #91, opened 2016-03-02
Adds ZipkinRule.receivedSpanCount and ZipkinRule.receivedSpanBytes

adriancole:
This adds ZipkinRule.receivedSpanCount and ZipkinRule.receivedSpanBytes
to help instrumentation projects tune the amount of data they report.

Using this, instrumentation projects can verify they are sending less
bytes, ex via compression or less data duplication. They can also verify
they aren't sending more spans they expect. For example, instrumentation
can easily bundle the same span multiple times in the same POST request.

### DIFF
--- a/zipkin-junit/src/test/java/zipkin/junit/ZipkinRuleTest.java
+++ b/zipkin-junit/src/test/java/zipkin/junit/ZipkinRuleTest.java
@@ -90,6 +90,19 @@ public class ZipkinRuleTest {
     assertThat(zipkin.httpRequestCount()).isEqualTo(2);
   }
 
+  /**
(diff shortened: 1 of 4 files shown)


zipkin-junit/src/test/java/zipkin/junit/ZipkinRuleTest.java
```diff
@@ -91,6 +91,14 @@ public void httpRequestCountIncrements() throws IOException {
   }
 
   @Test
+  public void receivedSpanCountIncrements() throws IOException {
+    postSpans(span);
```
adriancole: normally, a span can be reported twice: for client and server. However, there are bugs that happened where several updates went to the same span id. This counter can be used to cement in expectations.

PR #92, opened 2016-03-07
Adds CassandraSpanStore including zipkin-server support

### LINKED ISSUE #89: Implement CassandraSpanStore
Cassandra is a core zipkin backend. Many users will not be able to use this project until it is supported here.

adriancole:
This is a port of `CassandraSpanStore` from zipkin-scala, except notably
not including the "slice query" infrastructure.

Note: While `CassandraSpanStore.accept` doesn't return a `Future`, it is
still asynchronous inside.

`zipkin-server` has been extended to optionally support Cassandra using
the same environment variables as zipkin-scala. For example, you can
start a process using cassandra as simply as setting the `STORAGE_TYPE`:

```bash
STORAGE_TYPE=cassandra ./mvnw -pl zipkin-server spring-boot:run
```

This uses the `Repository` class from `zipkin-cassandra-core`, which
internally uses DataStax Java Driver 2.1. It is likely that we'll
refactor out this dependency, for a couple reasons. One is that DataStax
Java Driver 3 is current. Another is that `Repository` was designed for
"slice query" infrastructure that doesn't exist in this project: there
are places where logic would be simpler to address in raw CQL.

This was tested using zipkin-scala's integration tests, local integration
tests, running the server with spring boot, and processing data with
`zipkin-dependencies-spark`.

Fixes #89

### DIFF
--- a/zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
+++ b/zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
@@ -1,0 +1,194 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package zipkin.cassandra;
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.HostDistance;
+import com.datastax.driver.core.PoolingOptions;
+import com.datastax.driver.core.policies.DCAwareRoundRobinPolicy;
+import com.datastax.driver.core.policies.LatencyAwarePolicy;
+import com.datastax.driver.core.policies.RoundRobinPolicy;
+import com.datastax.driver.core.policies.TokenAwarePolicy;
+import com.google.common.collect.Sets;
+import com.google.common.net.HostAndPort;
+import java.net.InetSocketAddress;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import org.twitter.zipkin.storage.cassandra.ZipkinRetryPolicy;
+import zipkin.internal.Nullable;
+
+import static zipkin.internal.Util.checkNotNull;
+
+/** Configuration including defaults needed to run against a local Cassandra installation. */
+public final class CassandraConfig {
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static final class Builder {
+    private String keyspace = "zipkin";
+    private String contactPoints = "localhost";
+    private String localDc;
+    private int maxConnections = 8;
+    private boolean ensureSchema = true;
+    private String username;
+    private String password;
+    private int maxTraceCols = 100000;
+    private int spanTtl = (int) TimeUnit.DAYS.toSeconds(7);
+    private int indexTtl = (int) TimeUnit.DAYS.toSeconds(3);
+
+    /** Keyspace to store span and index data. Defaults to "zipkin" */
+    public Builder keyspace(String keyspace) {
+      this.keyspace = keyspace;
+      return this;
+    }
+
+    /** Comma separated list of hosts / IPs part of Cassandra cluster. Defaults to localhost */
+    public Builder contactPoints(String contactPoints) {
+      this.contactPoints = contactPoints;
+      return this;
+    }
+
+    /**
+     * Name of the datacenter that will be considered "local" for latency load balancing. When
+     * unset, load-balancing is round-robin.
+     */
+    public Builder localDc(@Nullable String localDc) {
+      this.localDc = localDc;
+      return this;
+    }
+
+    /** Max pooled connections per datacenter-local host. Defaults to 8 */
+    public Builder maxConnections(int maxConnections) {
+      this.maxConnections = maxConnections;
+      return this;
+    }
+
+    /**
+     * Ensures that schema exists, if enabled tries to execute script
+     * io.zipkin:zipkin-cassandra-core/cassandra-schema-cql3.txt. Defaults to true.
+     */
+    public Builder ensureSchema(boolean ensureSchema) {
+      this.ensureSchema = ensureSchema;
+      return this;
+    }
+
+    /** Will throw an exception on startup if authentication fails. No default. */
+    public Builder username(@Nullable String username) {
+      this.username = username;
+      return this;
+    }
+
+    /** Will throw an exception on startup if authentication fails. No default. */
+    public Builder password(@Nullable String password) {
+      this.password = password;
+      return this;
+    }
+
+    /**
+     * Spans have multiple values for the same id. For example, a client and server contribute to
+     * the same span id. When searching for spans by id, the amount of results may be larger than
+     * the ids. This defines a threshold which accommodates this situation, without looking for an
+     * unbounded number of results.
+     */
+    public Builder maxTraceCols(int maxTraceCols) {
+      this.maxTraceCols = maxTraceCols;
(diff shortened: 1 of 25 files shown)


zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanConsumer.java
```diff
@@ -0,0 +1,98 @@
...
+    this.repository = repository;
+    this.spanTtl = spanTtl;
+    this.indexTtl = indexTtl;
+  }
+
+  /**
+   * <p>Storing spans result in asynchronous operations to the backend repository. This
+   * implementation neither blocks on nor checks these futures, as the api doesn't require it.
+   */
+  void accept(Iterator<Span> spans) {
```
adriancole: maybe I need to return an aggregate future to avoid test race conditions.. (assuming Future.get means operation done with read-your-writes consistency..) I somehow remember @danchia needing to do this.

adriancole:
Early look time, cassandra folks @michaelsembwever @kristofa @danchia @yurishkuro @prat0318 @eirslett

yurishkuro:
what are "slice queries"?

adriancole:
the base class used in zipkin-scala
https://github.com/openzipkin/zipkin/blob/master/zipkin-common/src/main/scala/com/twitter/zipkin/storage/CollectAnnotationQueries.scala

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
```diff
@@ -0,0 +1,190 @@
...
+    int defaultPort = findConnectPort(contactPoints);
+    builder.addContactPointsWithPorts(contactPoints);
+    builder.withPort(defaultPort); // This ends up protocolOptions.port
+    if (username != null && password != null) {
+      builder.withCredentials(username, password);
+    }
+    builder.withRetryPolicy(ZipkinRetryPolicy.INSTANCE);
+    builder.withLoadBalancingPolicy(new TokenAwarePolicy(new LatencyAwarePolicy.Builder(
+        localDc != null
+            ? DCAwareRoundRobinPolicy.builder().withLocalDc(localDc).build()
```
danchia: Not sure what other people think, but I wonder if it's better to always use `DCAwareRoundRobinPolicy`.
Even if `localDC` isn't set, the java driver is supposed to infer the DC based on the first node contacted among `contactPoints`, which typically is set only to nodes in the same DC.

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanConsumer.java
danchia: For ease of testing with async APIs like this, it's definitely better to try and return a Future that completes when the underlying operations have all finished.

If the caller doesn't care, the caller can just ignore the returned future. For tests where we do, we can just block on the future completing.

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanConsumer.java
```diff
@@ -0,0 +1,106 @@
...
+      if (span.timestamp != null) {
+        for (ByteBuffer annotation : annotationKeys(span)) {
+          futures.add(repository.storeTraceIdByAnnotation(annotation, span.timestamp, span.traceId,
+              indexTtl));
+        }
+      }
+    }
+    if (futures.isEmpty()) return;
+    // TODO: getting futures to see if this helps w/ test flake in travis
+    Futures.getUnchecked(Futures.allAsList(futures));
```
danchia: This doesn't seem to agree with the javadoc of the function, which claims this implementation doesn't block or check on the operations finishing.

adriancole: >
> > +    // TODO: getting futures to see if this helps w/ test flake in travis
> > +    Futures.getUnchecked(Futures.allAsList(futures));
>
> This doesn't seem to agree with the javadoc of the function, which claims
> this implementation doesn't block or check on the operations finishing.
>
This was a test to see if blocking fixed the flake.. seems my comment
failed to make that point

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanConsumer.java
adriancole: I don't think we should change the api to return a Future because of a test. What would we do for a datastore that doesn't offer read-your-writes in that case? Implement a future that does  a blocking loop? How Future.cancel? Also, remember this interface is dependency-free. This means that it would be a java Future, not a guava ListenableFuture as the return type. This means no callback registration or composition.

Right now, the only use case for blocking on the future is to get read-your-writes during tests. This is very special-case. For example, collection endpoints are completely decoupled from query ones.. There's no use case for composing collection with query in the same thread, in other words.

For reasons above, I really don't like forcing things to return a Future, and it is premature to design an alternative (such as a callback that could be used in things like rx). That doesn't remove the travis flake problem, though.

To solve the travis-flake problem, I'll add a flag to block on futures, and document this as test-only. If somewhere down the line, we come up with a use case for async on the collector method (ex there's none currently in the scala impl either), then happy to solve it, but I doubt I'd solve it with Future.

Make sense?

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
adriancole: good question. this is just recreating the logic in zipkin-scala. I've no problem changing both to always use DCAwareRoundRobinPolicy.. @yurishkuro any thoughts on this?

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanConsumer.java
danchia: Yup that definitely makes sense. Especially exposing guava futures – definitely can be a pain to have Guava in the API layer.

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
yurishkuro: I've never tested DC-aware with null dc. If it works, +1

adriancole:
blocking on futures fixed test race conditions. opened an internal flag with lots of docs around rationale.

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
adriancole: https://github.com/openzipkin/zipkin/pull/1016

adriancole:
backfilled config tests and adjusted to DCAwareRoundRobinPolicy

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
adriancole: done (here)

michaelsembwever: The risk is that if it's null (localDC is not set), and the first contact node is in a distant datacenter, all connections will be to that distant datacenter.

michaelsembwever:
LGTM!

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraConfig.java
adriancole: That would be unlucky wouldn't it! I guess at least plain round robin would
have a chance of not always hitting remote

michaelsembwever: Yes. The LatencyAware around the round robin will then bring it back to just using the local datacenters nodes.

adriancole: Hmm so should we revert? (This is in both impls and I don't mind reverting
especially with the notes you've made)

michaelsembwever: It's not crucial. But it's a simple thing to do, and in lei of any other objections/experiences, I'd suggest the small commit changing the one line :-)

PR #95, opened 2016-03-14
Adds Kafka Transport

### LINKED ISSUE #70: Kafka Span Transport
Tracers in python, ruby, and java support a kafka transport. We need to implement this feature such that zipkin-java-servers can accept spans in the same way.

Requested by @ewhauser

> I'm looking at what exists in zipkin today because I'm looking to deploy a couple large installations of zipkin. We mostly work in Java, so zipkin-java-server makes sense for us. We also want to use HTTP as the transport mechanism because we will have some cross data center traffic.
I was looking into adding support for writing spans to Kafka, but there's a couple issues I need to figure out. 1) we probably need a way of differentiating between read and write span stores since Kafka would be the writer and another store would provide the reads 2) the interface for SpanStore currently takes an Iterator<Span> and I really want the raw byte[] so I don't have to serialize again before writing to Kafka

adriancole:
This adds a Kafka transport, which receives messages in the same way as
zipkin-scala, using the same environment variables.

Closes #70

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
     <module>benchmarks</module>
     <module>interop</module>
     <module>zipkin-spanstores</module>
+    <module>zipkin-transports</module>
     <module>zipkin-server</module>
   </modules>
 
@@ -41,7 +42,7 @@
 
     <moshi.version>1.1.0</moshi.version>
     <okio.version>1.6.0</okio.version>
-    <spring-boot.version>1.3.2.RELEASE</spring-boot.version>
+    <spring-boot.version>1.3.3.RELEASE</spring-boot.version>
     <zipkin-scala.version>1.35.0</zipkin-scala.version>
     <!-- MySQL connector is GPL, even if it has an OSS exception.
          https://www.mysql.com/about/legal/licensing/foss-exception/
@@ -142,6 +143,12 @@
       <dependency>
         <groupId>${project.groupId}</groupId>
         <artifactId>spanstore-jdbc</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>transport-kafka</artifactId>
         <version>${project.version}</version>
       </dependency>
 
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -58,21 +58,6 @@
       <artifactId>spring-boot-starter-actuator</artifactId>
     </dependency>
 
-    <!-- Trace api controller activity with Brave -->
-    <dependency>
-      <groupId>com.github.kristofa</groupId>
-      <artifactId>brave-spring-web-servlet-interceptor</artifactId>
-      <version>${brave.version}</version>
-      <optional>true</optional>
-    </dependency>
-
-    <dependency>
-      <groupId>com.github.kristofa</groupId>
-      <artifactId>brave-spancollector-scribe</artifactId>
-      <version>${brave.version}</version>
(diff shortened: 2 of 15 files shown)


zipkin-server/src/main/java/zipkin/server/ZipkinServerConfiguration.java
```diff
@@ -105,4 +107,33 @@ public Object postProcessAfterInitialization(Object bean, String beanName) {
...
+
+    @Override
+    public Object postProcessBeforeInitialization(Object bean, String beanName) {
+      return bean;
+    }
+
+    // Attempt to make Kafka listener registration conditional on kafka.getZookeeper() != null
+    @Override
+    public Object postProcessAfterInitialization(Object bean, String beanName) {
+      if (bean instanceof ZipkinSpanWriter && kafka.getZookeeper() != null) {
```
adriancole: @dsyer @marcingrzejszczak any ideas here? basically, I want to conditionally register a bean so that it can be closed. problem is that I can't find a `@ConditionalOnExpression` that can get me there, so I'm being (too) clever.

adriancole:
@kristofa @ewhauser if one of you would like to try this (for brave testing), pull this branch and run the following. Feedback is welcome!

`KAFKA_ZOOKEEPER=localhost:2081 ./mvnw -pl zipkin-server spring-boot:run`

pom.xml
```diff
@@ -41,7 +42,7 @@
 
     <moshi.version>1.1.0</moshi.version>
     <okio.version>1.6.0</okio.version>
-    <spring-boot.version>1.3.2.RELEASE</spring-boot.version>
+    <spring-boot.version>1.3.3.RELEASE</spring-boot.version>
```
adriancole: bump needed to dodge SLF4J version conflict in kafka

zipkin-server/pom.xml
```diff
@@ -108,6 +93,28 @@
       <optional>true</optional>
     </dependency>
 
+    <!-- Kafka transport -->
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>transport-kafka</artifactId>
+      <optional>true</optional>
```
adriancole: integration test still passes, so I think I've managed to make this really actually optional

zipkin-server/src/main/java/zipkin/server/ZipkinServerConfiguration.java
marcingrzejszczak: Here you have an example of a custom condition annotation - https://github.com/spring-cloud/spring-cloud-zookeeper/blob/master/spring-cloud-zookeeper-discovery/src/main/java/org/springframework/cloud/zookeeper/discovery/dependency/ConditionalOnDependenciesPassed.java and its implementation https://github.com/spring-cloud/spring-cloud-zookeeper/blob/master/spring-cloud-zookeeper-discovery/src/main/java/org/springframework/cloud/zookeeper/discovery/dependency/DependenciesPassedCondition.java#L37 . In the latter you can write whatever you need. Then you can put that annotation on your config

adriancole: Thanks for the tip!

adriancole: ok done!

zipkin-server/pom.xml
adriancole: definitely is optional. I screwed around with config and the integration test `zipkin-server/src/it/minimal-dependencies/`will trip if kafka classes are required to be loaded.

PR #96, opened 2016-03-15
Reverts default Cassandra host policy to RoundRobin

adriancole:
When changing loadbalancer policy, we didn't consider a scenario brought
up by @michaelsembwever:

DCAwareRoundRobinPolicy will by default prefer the first contact node's
datacenter. If that node happens to be in a remote datacenter, latency
will be high, and the `LatencyAwarePolicy` cannot correct it (since
local nodes will be filtered out.

The previous logic is better because eventhough `RoundRobinPolicy` can
select a remote datacenter, it also selects local ones: the
`LatencyAwarePolicy` will eventually correct to choose only the local
ones.

See https://github.com/openzipkin/zipkin-java/pull/92/files#r55933918


michaelsembwever:
:+1:

PR #98, opened 2016-03-16
Hosts zipkin-ui static assets on zipkin-server

adriancole:
This hosts the `zipkin-ui` static assets on zipkin-server as opposed to relying on an external process (zipkin-web)

By doing, so, this allows us the following:
* Single-jar deployment option
  * Simplest deployment of in-memory storage has no external dependencies
* The opportunity to use the standard port 8080
  * Since web assets are hosted locally, there's no port dodging anymore

Implementation note: this currently cherry-picks routes the client-side is
responsible for.

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -29,6 +29,7 @@
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
     <brave.version>3.5.0</brave.version>
+    <zipkin-ui.version>1.38.0</zipkin-ui.version>
     <start-class>zipkin.server.ZipkinServer</start-class>
     <maven-invoker-plugin.version>2.0.0</maven-invoker-plugin.version>
   </properties>
@@ -73,6 +74,14 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>zipkin</artifactId>
+    </dependency>
+
+    <!-- Static content for the web UI -->
+    <dependency>
+      <groupId>io.zipkin</groupId>
+      <artifactId>zipkin-ui</artifactId>
+      <version>${zipkin-ui.version}</version>
+      <optional>true</optional>
     </dependency>
 
     <!-- Cassandra backend -->
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -41,6 +41,12 @@ zipkin:
     lookback: ${QUERY_LOOKBACK:86400000}
   store:
     type: ${STORAGE_TYPE:mem}
+  # Values here are read by Zipkin UI's javascript at /config.json
+  ui:
+    # Default limit for Find Traces
+    query-limit: 10
+    # The value here becomes a label in the top-right corner
+    environment:
 server:
   port: ${QUERY_PORT:9411}
   compression:
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -1,5 +1,8 @@
 # zipkin-server
-The  receives spans via HTTP POST and respond to queries from zipkin-web.
+The hosts the Zipkin [Api](http://zipkin.io/zipkin-api/#/) and [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui).
+
+Span storage and transports are configurable. By default storage is
+in-memory and the http span transport (POST /spans endpoint) is enabled.
 
 Note that the server requires minimum JRE 8.
 
--- a/zipkin-server/src/main/java/zipkin/server/EnableZipkinServer.java
+++ b/zipkin-server/src/main/java/zipkin/server/EnableZipkinServer.java
@@ -24,7 +24,7 @@ import zipkin.server.brave.BraveConfiguration;
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Import({ZipkinServerConfiguration.class, BraveConfiguration.class, ZipkinQueryApiV1.class, ZipkinSpanWriter.class})
+@Import({ZipkinServerConfiguration.class, BraveConfiguration.class, ZipkinQueryApiV1.class, ZipkinUiConfiguration.class, ZipkinSpanWriter.class})
 public @interface EnableZipkinServer {
 
 }
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinServerProperties.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinServerProperties.java
@@ -23,6 +23,12 @@ class ZipkinServerProperties {
     return store;
   }
 
+  private Ui ui = new Ui();
+
+  public Ui getUi() {
+    return ui;
+  }
+
   static class Store {
     enum Type {
       cassandra, mysql, mem
@@ -36,6 +42,27 @@ class ZipkinServerProperties {
 
     public void setType(Type type) {
       this.type = type;
+    }
+  }
+
+  static class Ui {
+    private String environment;
+    private int queryLimit = 10;
+
+    public String getEnvironment() {
+      return environment;
+    }
+
+    public void setEnvironment(String environment) {
+      this.environment = environment;
+    }
+
+    public int getQueryLimit() {
+      return queryLimit;
+    }
+
+    public void setQueryLimit(int queryLimit) {
+      this.queryLimit = queryLimit;
     }
   }
 }
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinUiConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinUiConfiguration.java
@@ -1,0 +1,74 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.server;
+
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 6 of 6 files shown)


adriancole:
super thanks to the zipkin UI team @eirslett and @abesto 

I know that @dsyer will be very happy about this, since zipkin-web's file count makes it difficult to deploy in some environments.

zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -37,6 +37,12 @@ zipkin:
     lookback: ${QUERY_LOOKBACK:86400000}
   store:
     type: ${STORAGE_TYPE:mem}
+  # Values here are read by Zipkin UI's javascript at /config.json
+  web:
```
adriancole: note: this is a super fancy way to make these configurable..

The no-code way is to revert changes to `zipkin-server/src/main` and instead plop a `zipkin-server/src/main/resources/static/config.json`

```json
{
"environment": "",
"queryLimit": 10
}
```

eirslett:
It must serve HTTP endpoints in this order:
1) routes defined by zipkin-server
2) assets that exist in /static/
3) fallback to serving index.html

adriancole:
@spencergibb @marcingrzejszczak @spencergibb I've hunted around a bit, trying things like extending `WebMvcConfigurationSupport` to try and change priority. I'm not very experienced in this.. do you happen to know how to get the resolve behavior Eirik mentioned? Most notably, crossroads (in zipkin-ui javascript) owns the routes `traces/{id}` and `dependency`

zipkin-server/pom.xml
```diff
@@ -26,9 +26,23 @@
   <artifactId>zipkin-server</artifactId>
   <name>Zipkin Server</name>
 
+  <repositories>
```
dsyer: This needs to be in a profile (you can make it active by default using the .mvn configuration).

dsyer:
I'm not sure what you mean by "owns the routes". I haven't even looked at the JavaScript to see what you mean, but there is nothing in the server that provides content on those paths, so I'm unclear as to what the actual problem is.

Serving the zipkin-ui up as a big flat /static on the classpath is ugly, by the way, and not very embeddable because other jars now might have problems using that slot for static resources (it would be better IMO to put those assets in webjars-like form under /META-INF with a unique path to zipkin-ui).

eirslett:
> I'm not sure what you mean by "owns the routes".

The backend does no UI routing whatsoever. All it does is serve `index.html`, and all `index.html` does is load `app.min.js` and `app.min.css`.

When you `GET /dependencies` against the server, it should serve/fallback to serving `index.html`, since there is no `src/main/resources/dependencies` file. The JavaScript code looks at `window.location` to figure out what the UI should do.

abesto:
@dsyer We're in the unhappy situation where all the people working on this right now have little experience with web packages in the Java world, so your input on the /static thing is awesome. Can you give a specific example of the directory structure that'd make sense? My best guess: `/static/index.html` would go into `/META-INF/resources/zipkin-ui/index.html`. Does that sound about right?

If you can point me to relevant documentation, that'd be awesome.

eirslett:
Distributing zipkin-ui as a jar is a short-term solution, our longer-term goal is to distribute it as a .zip, or to npm (it's possible to autogenerate webjars from npm, I think?)
We want the UI to be independent of the implementation of the backend server. E.g. the backend could be an nginx server, in which case the META-INF/resources/... layout wouldn't make sense, since it's java/servlet-specific.

With gradle, it should be possible to extract whatever is inside the jar (or zip) file, to build/.../META-INF/resources/... so in zipkin-java we can put it wherever we want to, without conflicts.

eirslett:
> I've hunted around a bit, trying things like extending WebMvcConfigurationSupport to try and change priority.

Couldn't we just override the 404/route-not-found-handler, and serve index.html?

> not very embeddable because other jars now might have problems using that slot for static resources

I discussed with @abesto - We could rename "static" in the zipkin-ui distribution to "zipkin-ui", to prevent resources from conflicting.

dsyer:
> Distributing zipkin-ui as a jar is a short-term solution

Regardless of the other long term objectives, I think this should remain one of them. If the layout is sensible and constructed so it doesn't clash with anything else then it will be extremely valuable.

@abesto the Servlet spec makes a special place for /META-INF/resources, so any jar that carries resources there is in good shape to be embedded. I don't want to say that only Servlet apps should be able to consume it, but we might as well use their convention to make it as widely consumable as possible. The [webjars format](http://www.webjars.org/) basically builds on top of that, so `META-INF/resources/webjars/zipkin-ui/1.30.2` would be the best location for the resources currently in "static" (IMO). Alternatively we could punt on that and ask webjars to make their own re-packaging of our resources, but that seems overly complicated when we all basically want the same thing at the end of the day.

> When you GET /dependencies against the server, it should serve/fallback to serving index.html

OK, that's not so hard. But you need an active server component to do that. In this case it could be a Spring `@RequestMapping`. I'll make a proposal - the only thing that makes it slightly less than straightforward is that I'd like to support the use of the [webjars locator](https://github.com/webjars/webjars-locator) so we don't need to know the version in order to code the index.html.

eirslett:
> If the layout is sensible and constructed so it doesn't clash with anything else then it will be extremely valuable.

I mentioned that we could rename the directory name "static" to "zipkin-ui" to prevent clashes. Then we'll have a KISS packaging solution. It's quite simple to use from nginx too, since they can extract the .jar file like it's a .zip file (which it basically is) and then all the resources are there, in the zipkin-ui directory.
Would that be enough to make things work for now? Is there a reason why /META-INF/resources is especially embeddable? Then we won't need to involve webjars at all, and not the webjars locator either - the version number won't be in the assets' location.

dsyer:
I guess we can work with whatever you can provide. The more unique the path is the better, but if you don't want to support Servlet apps directly, we can make an adapter.

adriancole:
Thanks for all the notes, folks. Enlightening, particularly that the root concern is about forwarding to index.html vs returning a 404 for the special routes.

eirslett:
This way of serving the same html "wrapper container" for all routes is a basic requirement for modern client-side/single-page applications that leverage the JavaScript history API. It's supposed to be really simple to configure on the backend side. Finatra does it like this: https://github.com/openzipkin/zipkin/blob/master/zipkin-query/src/main/scala/com/twitter/zipkin/query/ZipkinQueryController.scala#L113

zipkin-server/pom.xml
adriancole: this is temporary, and won't be merged

adriancole:
OK everything works now. Instead of using a global redirect handler, I've cherry-picked routes for now.

zipkin-server/src/main/java/zipkin/server/ZipkinUI.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * <p>There are a couple ways to address this from the server-side. One is to match all unknown
+ * paths, and instead of returning 404, route them back to index.html. Another is to be cherry-pick
+ * well-known paths that the UI needs, and forward those explicitly. The current implementation uses
+ * the cherry-pick approach until a better way is determined.
+ *
+ * <p>See https://github.com/openzipkin/zipkin/pull/1045 for current progress on this topic.
+ */
+@RestController
+public class ZipkinUI {
```
adriancole: @dsyer @eirslett @abesto this was the simplest way I could figure out to get the current codebase to operate. Note that this works and requires no other configuration or special packaging. This also doesn't interfere with other routes that should return a 404.

I'm open to ideas, but I like that this is really low effort, and doesn't front-load having folks understand webjar (which includes requiring frameworks to parse a version location in the tree). I don't think the scala side knows webjar, for example.

dsyer:
I added a small change there to make the UI handler a bit less noisy. Can we change the class name from `ZipkinUI` to `ZipkinUi`?

adriancole:
Sure! Note there's some upstream changes to the org before merge. Ex this
isn't in 0.8.0 (just released) hopefully, 0.9.0 includes this feature (in a
couple days or less)

eirslett:
We should probably wait for the discussions abaout packaging in zipkin-ui to settle first :wink: 
zipkin surely cannot be the only project with this requirement (https://github.com/openzipkin/zipkin-java/pull/98#issuecomment-197252044), we should investigate building something more generic, on the library level, that spring boot could merge upstream? Basically re-implement [fileOrIndex](https://github.com/twitter/finatra/blob/458c6b639c3afb4e29873d123125eeeb2b02e2cd/http/src/main/scala/com/twitter/finatra/http/response/ResponseBuilder.scala#L321) in spring boot?

dsyer:
The "fileOrIndex" feature is kind of what `ResourceHttpRequestHandler` does. I'll have to ask the MVC guys how they would configure it to do what we need. As you say, it's pretty common to need that, so if it isn't easy it should be.

abesto:
Quick update: @adriancole, @eirslett and I have reached consensus on zipkin-ui packaging. Long story short: `fileOrIndex` is the only way we'll support hosting zipkin-ui, at least to start with. See https://github.com/openzipkin/zipkin/pull/1050 if you want gritty details.

zipkin-server/src/main/java/zipkin/server/ZipkinUI.java
```diff
@@ -50,13 +50,9 @@
     return server.getWeb();
   }
 
-  @RequestMapping(value = "/traces/{id}", method = RequestMethod.GET)
-  public ModelAndView forwardTraceEndpoint(ModelMap model) {
-    return new ModelAndView("forward:/index.html", model);
+  @RequestMapping(value = { "/traces/{id}", "/dependency"}, method = RequestMethod.GET)
```
adriancole: nice!

zipkin-server/src/main/java/zipkin/server/ZipkinUI.java
```diff
@@ -50,13 +50,9 @@
...
+  public String forwardTraceEndpoint() {
```
adriancole: this works unless you have "/" in the values. In that case, it literally returns the string below. so had to revert..

adriancole:
OK bumped against latest zipkin upstream. I've made the UI configuration dependent on the zipkin-ui content being in the classpath.

adriancole:
OK this is pinned to upstream 1.38 as opposed to a snapshot. Still works like a charm!

adriancole:
will cut 0.9 with this change, so we can give it a spin. Further polishing can be released quickly, like 0.9.1

adriancole:
0.9 is on the way to maven central now. will update docker next

PR #104, opened 2016-03-18
Allows transport-kafka to consume JSON encoded spans

adriancole:
By allowing JSON encoded span lists as Kafka message values, we enable
users to send the same payload as they can over the http transport. It
also allows a very easy debug story, such as sending test spans via CLI.

This implementation peeks at the first byte in the Kafka message, and
attempts to decode json, if it is a '[', or decimal 91. When using
TBinaryProtocol (thrift) encoding, we know that the first byte will be
the Type, which is decimal <=16, so there's no risk of clash.

See https://github.com/openzipkin/zipkin/pull/1052

### DIFF
--- a/zipkin-transports/kafka/README.md
+++ b/zipkin-transports/kafka/README.md
@@ -1,16 +1,31 @@
 # transport-kafka
 This transport polls a Kafka 8.2.2+ topic for messages that contain
-TBinaryProtocol big-endian encoded lists of spans. These spans are
-pushed to a span consumer.
+a list of spans in json or TBinaryProtocol big-endian encoding. These
+spans are pushed to a span consumer.
 
 `zipkin.kafka.KafkaConfig` includes defaults that will operate against a
 Kafka topic advertised in Zookeeper.
 
+### Encoding spans into Kafka messages
+The message's binary data includes a list of spans. Supported encodings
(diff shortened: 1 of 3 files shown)


adriancole:
tested with the following things..

In one tab, start up kafka and post a span list to it
```bash
$ docker run -d -p 2181:2181 -p 9092:9092 --env ADVERTISED_HOST=192.168.99.100  --env AUTO_CREATE_TOPICS=true spotify/kafka
$ kafka-console-producer.sh --broker-list 192.168.99.100:9092 --topic zipkin
[{"traceId":"1","name":"bang","id":"2","timestamp":1234,"binaryAnnotations":[{"key":"lc","value":"bamm-bamm","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]
```

In a different tab, run the server
```bash
$ KAFKA_ZOOKEEPER=192.168.99.100:2181 ./mvnw -pl zipkin-server spring-boot:run
```

Finally, curl and see the span arrived!
```bash
$ curl -s localhost:9411/api/v1/trace/1
[{"traceId":"0000000000000001","name":"bang","id":"0000000000000002","timestamp":1234,"annotations":[],"binaryAnnotations":[{"key":"lc","value":"bamm-bamm","endpoint":{"serviceName":"flintstones","ipv4":"127.0.0.1"}}]}]
```

PR #105, opened 2016-03-18
Exposes CASSANDRA_SPAN_TTL and CASSANDRA_INDEX_TTL variables

adriancole:
Span and Index Time-to-lives were already configurable, but not via properties. This fixes that.

   * `CASSANDRA_SPAN_TTL`: Time-to-live in seconds for span data. Defaults to 604800 (7 days)
   * `CASSANDRA_INDEX_TTL`: Time-to-live in seconds for index data. Defaults to 259200 (3 days)

By adding the above, we remain compatible with the scala query and collector services.

See https://github.com/openzipkin/zipkin/pull/1055


adriancole:
tested by enabling debug logging and passing the variables to the commandline like so:

```bash
STORAGE_TYPE=cassandra CASSANDRA_SPAN_TTL=1 CASSANDRA_INDEX_TTL=2 ./mvnw -pl zipkin-server spring-boot:run
```

PR #109, opened 2016-03-19
Backfills missing unit tests for dependency linker

adriancole:
Problems around dependencies come in two major forms.
* The storage layer, which build DependencyLinkSpans
* The DependencyLinker, which processes DependencyLinkSpans

Before, there wasn't an easy way to separate these problems:
DependenciesTest existed, but was invoked from the storage layer.

By backfilling DependencyLinker tests, problems related to dependency
links can be resolved closer to the code responsible for them.

### DIFF
--- a/zipkin/src/main/java/zipkin/SpanStore.java
+++ b/zipkin/src/main/java/zipkin/SpanStore.java
@@ -80,6 +80,10 @@ public interface SpanStore extends SpanConsumer {
    * the original endTs, even when bucketed. Using the daily example, if endTs was 11pm and lookback
    * was 25 hours, the implementation would query against 2 buckets.
    *
+   * <p>Some implementations parse {@link zipkin.internal.DependencyLinkSpan} from storage and call
+   * {@link zipkin.internal.DependencyLinker} to aggregate links. The reason is certain graph logic,
+   * such as skipping up the tree is difficult to implement as a storage query.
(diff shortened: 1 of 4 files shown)


zipkin/src/main/java/zipkin/SpanStore.java
```diff
@@ -80,6 +80,10 @@
    * the original endTs, even when bucketed. Using the daily example, if endTs was 11pm and lookback
    * was 25 hours, the implementation would query against 2 buckets.
    *
+   * <p>Some implementations parse {@link zipkin.internal.DependencyLinkSpan} from storage and call
```
adriancole: cc @anuraaga @eirslett This is pretty important regardless of whether dependency links cached in a table or not.

zipkin/src/main/java/zipkin/internal/DependencyLinker.java
```diff
@@ -78,7 +78,7 @@ public void putTrace(Iterator<DependencyLinkSpan> spans) {
         }
         parent = parent.parent();
       }
-      if (client == null) continue; // skip if no ancestors were servers
+      if (client == null || server == null) continue; // skip if no ancestors were servers
```
adriancole: what everyone wishes for :) the act of backfilling tests finds a bug!

PR #113, opened 2016-03-21
Ports all remaining interop tests from zipkin-scala 1.38 to java

### LINKED ISSUE #110: Port all interop tests from zipkin-scala 1.38 to java
We've ported some tests from scala to java, as many maintainers cannot debug them otherwise. We are still missing some important tests, such as ensuring span.duration is exported by query time

See https://github.com/openzipkin/zipkin/pull/1057

adriancole:
This is the last set of interop tests to port, particularly around
the dependency link functionality.

Fixes #110

### DIFF
--- a/zipkin/src/test/java/zipkin/DependenciesTest.java
+++ b/zipkin/src/test/java/zipkin/DependenciesTest.java
@@ -13,16 +13,22 @@
  */
 package zipkin;
 
-import java.util.Calendar;
-import java.util.GregorianCalendar;
-import java.util.Iterator;
 import java.util.List;
-import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import org.junit.Before;
 import org.junit.Test;
+import zipkin.internal.ApplyTimestampAndDuration;
+import zipkin.internal.Dependencies;
 
 import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
 import static org.assertj.core.api.Assertions.assertThat;
+import static zipkin.Constants.CLIENT_ADDR;
+import static zipkin.Constants.CLIENT_RECV;
+import static zipkin.Constants.CLIENT_SEND;
+import static zipkin.Constants.SERVER_ADDR;
+import static zipkin.Constants.SERVER_RECV;
+import static zipkin.Constants.SERVER_SEND;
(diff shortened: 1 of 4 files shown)


adriancole:
cc @anuraaga @marcingrzejszczak the scala test still run, but the java ones should fail for the same reasons, and be easier to debug

zipkin/src/test/java/zipkin/DependenciesTest.java
```diff
@@ -41,35 +47,330 @@
...
+            .addAnnotation(Annotation.create(1445136540397049L, CLIENT_RECV, twoPort3002))
+            .build()
+    );
+    processDependencies(trace);
+
+    long traceDuration = trace.get(0).duration;
+
+    assertThat(
+        store.getDependencies((trace.get(0).timestamp + traceDuration) / 1000, traceDuration / 1000)
+    ).containsExactly(
```
anuraaga: The scala version uses theSameElementsAs, which doesn't check order while this PR uses containsExactly which does. Should we change these to containsOnly?

anuraaga: For reference, scala
https://github.com/openzipkin/zipkin/blob/79161d905f5aafdfa3aa7d8c4a1acb5ddc69d66f/zipkin-common/src/test/scala/com/twitter/zipkin/storage/DependencyStoreSpec.scala#L128

zipkin/src/test/java/zipkin/DependenciesTest.java
```diff
@@ -41,35 +47,330 @@
...
+
+  /**
+   * The primary annotation used in the dependency graph is [[Constants.SERVER_RECV]]
+   */
+  @Test
+  public void getDependenciesMultiLevel() {
+    processDependencies(trace);
+
+    assertThat(store.getDependencies(today + 1000L, null))
+        .isEqualTo(dep.links);
```
anuraaga: I think isEqualTo has the same problem

zipkin/src/test/java/zipkin/DependenciesTest.java
adriancole: afaik there's no documented order for links, so feel free to relax to
contains only.

PR #117, opened 2016-03-22
Updates to zipkin-ui 1.39, notably addressing cache concerns

adriancole:
zipkin-ui includes some cache-related changes to make the experience
more responsive. It also introduces "Start time", which allows users
to control how far back to search for traces.

This includes a hard-coded cache policy, consistent with zipkin-scala.

* 1 minute for index.html
* 10 minute for /config.json
* 365 days for hashed resources (ex /app-e12b3bbb7e5a572f270d.min.js)

Since index.html links to hashed resource names, any change to it will
orphan old resources. That's why hashed resource age can be 365 days.

See #1062

### DIFF
--- a/zipkin-server/pom.xml
+++ b/zipkin-server/pom.xml
@@ -29,7 +29,7 @@
   <properties>
     <main.basedir>${project.basedir}/..</main.basedir>
     <brave.version>3.5.0</brave.version>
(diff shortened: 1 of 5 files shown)


adriancole:
this is waiting on a new version of zipkin-ui which hashes all resources except the index

adriancole:
reworking this to match upstream logic

Issue #111, opened 2016-03-21
Make spanstore interfaces asynchronous


anuraaga:
I have a use case for querying span stores in an asynchronous server and the blocking API makes me sad knowing most stores support full async.

In the note on BLOCK_ON_FUTURES, I noticed the decision on picking a future spec was punted but shall we get back to it?

Some options

1) RxJava Observable. In another issue I noticed people are already onboard with RxJava so this seems like the least controversial pick.

2) CompletableFuture. Forces span stores to Java8. There seems to be interest in keeping to Java7 for broader support but wonder whether there are actually use cases for slotting these into legacy code.

3) ListenableFuture. Requires Guava, which may be too large, though personally find working on pre-java8 code without it impossible. I took much advantage of ElasticSearch's dependency on it in my spanstore to preserve my sanity :)

4) Callback. Callbacks are gross, but lightest weight.

I like any of the non-callback solutions since converting between futures is easy :)

adriancole:
>
> I have a use case for querying span stores in an asynchronous server and
> the blocking API makes me sad knowing most stores support full async.
>
well I wouldn't say most at the moment. JDBC isn't async, and in-memory is.
Cassandra indeed returns guava futures, though, and with ElasticSearch it
become a tie :).

> In the note on BLOCK_ON_FUTURES, I noticed the decision on picking a
> future spec was punted but shall we get back to it?
>
So block on futures is about the store use-case. I think your use case is
for query, particularly using the SpanStore api directly.

The most important thing we can do is itemize what the minimum goal is, as
there are so many things people mean when they say async. Ex. are you
looking to defer the result or compose it?

> 1) RxJava Observable. In another issue I noticed people are already
> onboard with RxJava so this seems like the least controversial pick.
>
agreed that this is the first that we spoke about. bear in mind that
there's some activity on 2.x, which is a sad reality the moment we start
choosing async composition libraries.

> 2) CompletableFuture. Forces span stores to Java8. There seems to be
> interest in keeping to Java7 for broader support but wonder whether there
> are actually use cases for slotting these into legacy code.
>
well it depends on what the async use case is. if it isn't a composition
one, regular future could work (even if not a happy one).

> 3) ListenableFuture. Requires Guava, which may be too large, though
> personally find working on pre-java8 code without it impossible. I took
> much advantage of ElasticSearch's dependency on it in my spanstore to
> preserve my sanity :)
>
There are indeed a family libraries that expose guava, but this like other
[...]

PR #119, opened 2016-03-22
Add a GuavaSpanStore interface which exposes span store methods as Li…

anuraaga:
…stenableFuture, and an adaptor to the normal SpanStore. Migrate ElasticsearchSpanStore to it.

For #111. I would port cassandra in a following PR.

### DIFF
--- a/interop/src/test/java/zipkin/elasticsearch/ElasticsearchScalaDependencyStoreTest.java
+++ b/interop/src/test/java/zipkin/elasticsearch/ElasticsearchScalaDependencyStoreTest.java
@@ -24,6 +24,7 @@ import zipkin.InMemorySpanStore;
 import zipkin.SpanStore;
 import zipkin.interop.ScalaDependencyStoreAdapter;
 import zipkin.interop.ScalaSpanStoreAdapter;
+import zipkin.spanstore.guava.BlockingGuavaSpanStore;
 
 import static zipkin.internal.Util.midnightUTC;
 
@@ -36,7 +37,7 @@ public class ElasticsearchScalaDependencyStoreTest extends DependencyStoreSpec {
   }
 
   public DependencyStore store() {
-    return new ScalaDependencyStoreAdapter(spanStore);
+    return new ScalaDependencyStoreAdapter(new BlockingGuavaSpanStore(spanStore));
   }
 
   @Override
--- a/interop/src/test/java/zipkin/elasticsearch/ElasticsearchTestGraph.java
+++ b/interop/src/test/java/zipkin/elasticsearch/ElasticsearchTestGraph.java
@@ -15,6 +15,7 @@ package zipkin.elasticsearch;
 
 import org.elasticsearch.client.transport.NoNodeAvailableException;
 import org.junit.AssumptionViolatedException;
+import zipkin.spanstore.guava.BlockingGuavaSpanStore;
 
 enum ElasticsearchTestGraph {
   INSTANCE;
@@ -23,7 +24,8 @@ enum ElasticsearchTestGraph {
 
   static {
     // Avoid race-conditions in travis by forcing read-your-writes consistency.
-    ElasticsearchSpanConsumer.BLOCK_ON_FUTURES = true;
+    BlockingGuavaSpanStore.BLOCK_ON_ACCEPT = true;
+    ElasticsearchSpanConsumer.FLUSH_ON_WRITES = true;
   }
 
   private AssumptionViolatedException ex;
--- a/zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticListenableFuture.java
+++ b/zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticListenableFuture.java
@@ -1,0 +1,37 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package zipkin.elasticsearch;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.ListenableActionFuture;
+
+class ElasticListenableFuture {
+
+  static <T> ListenableFuture<T> of(ListenableActionFuture<T> elasticFuture) {
+    final SettableFuture<T> future = SettableFuture.create();
+    elasticFuture.addListener(new ActionListener<T>() {
+      @Override public void onResponse(T t) {
+        future.set(t);
+      }
+
+      @Override public void onFailure(Throwable e) {
+        future.setException(e);
+      }
+    });
+    return future;
+  }
+}
--- a/zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
+++ b/zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
@@ -14,11 +14,15 @@
 package zipkin.elasticsearch;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.google.common.net.HostAndPort;
+import com.google.common.util.concurrent.AsyncFunction;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
@@ -62,12 +66,13 @@ import zipkin.Codec;
 import zipkin.DependencyLink;
 import zipkin.QueryRequest;
 import zipkin.Span;
-import zipkin.SpanStore;
 import zipkin.internal.CorrectForClockSkew;
 import zipkin.internal.MergeById;
 import zipkin.internal.Nullable;
 import zipkin.internal.Util;
+import zipkin.spanstore.guava.GuavaSpanStore;
 
+import static com.google.common.util.concurrent.Futures.immediateFuture;
 import static org.elasticsearch.index.query.QueryBuilders.boolQuery;
 import static org.elasticsearch.index.query.QueryBuilders.matchAllQuery;
 import static org.elasticsearch.index.query.QueryBuilders.nestedQuery;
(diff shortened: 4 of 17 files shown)


interop/src/test/java/zipkin/elasticsearch/ElasticsearchScalaDependencyStoreTest.java
```diff
@@ -36,7 +37,7 @@ public static void setupDB() {
   }
 
   public DependencyStore store() {
-    return new ScalaDependencyStoreAdapter(spanStore);
+    return new ScalaDependencyStoreAdapter(new BlockingGuavaSpanStore(spanStore));
```
adriancole: ps another thing :) we could more directly implement scala adapter from a guava span store!

interop/src/test/java/zipkin/elasticsearch/ElasticsearchTestGraph.java
```diff
@@ -23,7 +24,8 @@
 
   static {
     // Avoid race-conditions in travis by forcing read-your-writes consistency.
-    ElasticsearchSpanConsumer.BLOCK_ON_FUTURES = true;
+    BlockingGuavaSpanStore.BLOCK_ON_FUTURES = true;
+    ElasticsearchSpanConsumer.FLUSH_ON_WRITES = true;
```
adriancole: thanks for making this setting. I would have suggested it, but you beat me to it!

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticListenableFuture.java
```diff
@@ -0,0 +1,38 @@
...
+package zipkin.elasticsearch;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.SettableFuture;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.ListenableActionFuture;
+
+class ElasticListenableFuture {
+
+  static <T> ListenableFuture<T> of(ListenableActionFuture<T> elasticFuture) {
```
adriancole: sweet I was hoping it would be this simple

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
```diff
@@ -164,40 +169,56 @@ public ElasticsearchSpanStore(ElasticsearchConfig config) {
...
-      return Collections.emptyList();
-    }
-    List<Long> traceIds = new ArrayList<>();
-    for (Terms.Bucket bucket : traceIdsAgg.getBuckets()) {
-      traceIds.add(Util.lowerHexToUnsignedLong(bucket.getKeyAsString()));
-    }
-    return getTracesByIds(traceIds, indices);
+    return Futures.transformAsync(
+        ElasticListenableFuture.of(elasticRequest.execute()),
+        new AsyncFunction<SearchResponse, List<List<Span>>>() {
```
adriancole: heh probably nearing the limit of indent level vs extract static class ;)

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
```diff
@@ -164,40 +169,56 @@ public ElasticsearchSpanStore(ElasticsearchConfig config) {
...
+        });
   }
 
-  @Override public List<Span> getTrace(long id) {
-    return Iterables.getFirst(getTracesByIds(ImmutableList.of(id), indexNameFormatter.catchAll()),
-        null);
+  @Override public ListenableFuture<List<Span>> getTrace(long id) {
+    return Futures.transform(
+        getTracesByIds(ImmutableList.of(id), indexNameFormatter.catchAll()),
+        new Function<List<List<Span>>, List<Span>>() {
```
adriancole: I bet you hate me for mentioning java 7, right? :P

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
```diff
@@ -164,40 +169,56 @@ public ElasticsearchSpanStore(ElasticsearchConfig config) {
...
     SearchRequestBuilder elasticRequest = client.prepareSearch(indexNameFormatter.catchAll())
         .setTypes(ElasticsearchConstants.SPAN)
       .setQuery(termQuery("traceId",String.format("%016x", traceId)));
-    SearchResponse response = elasticRequest.execute().actionGet();
-    ImmutableList.Builder<Span> trace = ImmutableList.builder();
-    for (SearchHit hit : response.getHits()) {
-      trace.add(Codec.JSON.readSpan(hit.getSourceRef().toBytes()));
-    }
-    return trace.build();
+    return Futures.transform(
```
adriancole: I noticed that this combination of transform, ElasticListenableFuture.of happens quite a bit. wdyt about `ElasticListenableFuture.transform(elastic, fn)`?

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
```diff
@@ -291,20 +331,26 @@ public ElasticsearchSpanStore(ElasticsearchConfig config) {
...
+            Terms namesAgg = response.getAggregations().get("name_agg");
+            if (namesAgg == null) {
+              return Collections.emptyList();
+            }
+            ImmutableList.Builder<String> spanNames = ImmutableList.builder();
+            for (Terms.Bucket bucket : namesAgg.getBuckets()) {
+              spanNames.add(bucket.getKeyAsString());
+            }
+            return spanNames.build();
+          }
```
adriancole: one side effect of anonymous vs inner static class is that the latter looks better in a stack trace (assuming it has a nice name), and is also easier to test. I'd consider extracting classes for callbacks that have more than a few lines.

adriancole: this particular one could go either way, since is isn't doing much.

zipkin-spanstores/guava/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>zipkin-spanstores</artifactId>
+    <groupId>io.zipkin.java</groupId>
+    <version>0.9.4-SNAPSHOT</version>
```
adriancole: promise I won't release before this merges :P

zipkin-spanstores/guava/src/main/java/zipkin/spanstore/guava/GuavaSpanStore.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * An interface that is equivalent to {@link zipkin.SpanStore} but exposes methods as
+ * {@link ListenableFuture} to allow asynchronous composition.
+ *
+ * @see zipkin.SpanStore
+ */
+public interface GuavaSpanStore extends GuavaSpanConsumer {
+
+  /**
```
adriancole: I'd use links rather than c&p the signatures, as this can be quite drifty.

ex.
```java
/** Like {@link SpanStore#accept}, except the result is a {@link ListenableFuture} */
```

zipkin-spanstores/guava/src/test/java/zipkin/spanstore/guava/BlockingGuavaSpanStoreTest.java
```diff
@@ -0,0 +1,201 @@
...
+    QueryRequest request = new QueryRequest.Builder("service").endTs(1000L).build();
+    when(delegate.getTraces(request)).thenReturn(immediateFuture(traces));
+    assertThat(spanStore.getTraces(request)).containsExactlyElementsOf(traces);
+  }
+
+  @Test
+  public void getTraces_exception() {
+    QueryRequest request = new QueryRequest.Builder("service").endTs(1000L).build();
+    when(delegate.getTraces(request)).thenThrow(new IllegalStateException("failed"));
+    thrown.expect(IllegalStateException.class);;
```
adriancole: great. I was wondering about that

adriancole:
cool beans. made some style notes, but looks good!

interop/src/test/java/zipkin/elasticsearch/ElasticsearchScalaDependencyStoreTest.java
anuraaga: Hmm - using zipkin-java spanstores from zipkin-scala does seem like an interesting Future (pun intended).

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
anuraaga: At first yes ;) But while writing this I remembered how much better IDEs (Intellij) are at dealing with anonymous classes than lambda. Most of my lambdas tend to be completely red until the very last character fixes the type inference, which is really annoying...

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
anuraaga: Cool idea, I played with it, but the amount of code saved didn't seem like it was worth introducing "2 ways to do one thing", so I decided against, let me know if this makes sense.

zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
anuraaga: Yeah this and one more barely passed my threshold, but seems to have not passed yours which is completely reasonable. Went ahead and extracted a couple more private methods.

Note, I tend to prefer an anonymous class that calls a private method (in java8 would just be a method reference) because it feels more like code, while defining inner static classes that have no state feels off to me. But admittedly just personal preference.

zipkin-spanstores/guava/src/main/java/zipkin/spanstore/guava/GuavaSpanStore.java
anuraaga: Great idea, fixed this and GuavaSpanConsumer.

anuraaga:
Thanks for the review, addressed comments

zipkin-spanstores/guava/src/main/java/zipkin/spanstore/guava/GuavaSpanConsumer.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * An interface that is equivalent to {@link zipkin.SpanConsumer} but exposes methods as
+ * {@link ListenableFuture} to allow asynchronous composition.
+ */
+// @FunctionalInterface
+public interface GuavaSpanConsumer {
+
+  /**
+   * Version of {@link zipkin.SpanStore#accept} that returns a {@link ListenableFuture}.
```
adriancole: supernit that I'll sort on the way in: s/SpanStore/SpanConsumer

zipkin-spanstores/guava/src/main/java/zipkin/spanstore/guava/GuavaSpanStore.java
```diff
@@ -0,0 +1,65 @@
...
+   * Version of {@link zipkin.SpanStore#accept} that returns {@link ListenableFuture}.
+  @Override
```
adriancole: oops! fixing on the way in

adriancole:
closed via 818ea2f99e9428b4bb6d41fbffee215e3634f92c

note I reworded the commit to not wrap and also add a little more context

PR #121, opened 2016-03-23
Define elasticsearch hosts property as a List<String> instead of comm…

anuraaga:
…a-separated list String. spring-boot handles the conversion correctly, even for environment variables defined as a comma-separated list.


anuraaga:
@shakuzen Thanks for the tip!

adriancole:
sweet

PR #124, opened 2016-03-24
Makes it possible for all span stores to implement AsyncSpanStore

### LINKED ISSUE #111: Make spanstore interfaces asynchronous

adriancole:
In a previous pull request, Elasticsearch implemented AsyncSpanStore.

This completes the work, with the following summary:

* `AsyncToScalaSpanStoreAdapter` can port an AsyncSpanStore to the scala SpanStore 
* `BlockingToAsyncSpanStoreAdapter` lets (the blockin) SpanStore implement AsyncSpanStore
  * `JDBCSpanStore` is blocking and can wrap with ^^
* `GuavaToAsyncSpanStoreAdapter` lets Guava implement AsyncSpanStore
  * `CassandraSpanStore` and `ElasticSearchSpanStore` now extend ^^
* `InMemorySpanStore` now directly implements AsyncSpanStore
  * callbacks are issued on the calling thread as opposed to defering

Fixes #111


adriancole:
@anuraaga massive refactor, but I think this will help us stabalize. I have some tests to fill and polishing, but since you're such a good reviewer, thought I'd give you a ping.

interop/src/main/java/zipkin/interop/CallbackWithPromise.java
```diff
@@ -0,0 +1,33 @@
...
+ * the License.
+ */
+package zipkin.interop;
+
+import com.twitter.util.Promise;
+import com.twitter.util.Promise$;
+import zipkin.async.Callback;
+import zipkin.internal.Nullable;
+
+abstract class CallbackWithPromise<C, P> implements Callback<C> {
```
adriancole: pulled out because there are two scala types DependencyStore and SpanStore.

interop/src/main/java/zipkin/interop/ScalaSpanStoreAdapter.java
```diff
@@ -161,4 +174,42 @@ public void close() {
       return null;
     }
   }
+
+  static Future<Seq<List<Span>>> collectNotNullAndSort(java.util.List<Future<List<Span>>> result) {
```
adriancole: In the future, we might want to move this stuff out of zipkin-java as it is very complex to code scala in java. For example, if this moved to the scala project, it could be written in scala :P

interop/src/main/java/zipkin/interop/ScalaSpanStoreAdapter.java
```diff
@@ -18,36 +18,48 @@
...
- * Adapts {@link SpanStore} to a scala {@link com.twitter.zipkin.storage.SpanStore} in order to test
- * against its {@link com.twitter.zipkin.storage.SpanStoreSpec} for interoperability reasons.
+ * Adapts {@link AsyncSpanStore} to a scala {@link com.twitter.zipkin.storage.SpanStore} in order to
+ * test against its {@link com.twitter.zipkin.storage.SpanStoreSpec} for interoperability reasons.
  *
  * <p/> This implementation uses thrift TBinaryProtocol to ensure structures are compatible.
  */
 public final class ScalaSpanStoreAdapter extends com.twitter.zipkin.storage.SpanStore {
-  private final SpanStore spanStore;
+  private final AsyncSpanStore spanStore;
```
adriancole: big change here is that everything moved from blocking and returning an immediate future, to actually deferring results

zipkin-server/src/main/java/zipkin/server/ZipkinServerConfiguration.java
```diff
@@ -151,7 +151,7 @@ public Object postProcessAfterInitialization(Object bean, String beanName) {
           .hosts(elasticsearch.getHosts())
           .index(elasticsearch.getIndex())
           .build();
-      return new BlockingSpanStoreAdapter(new ElasticsearchSpanStore(config));
```
adriancole: SpanStore is blocking, and after adding more adapters, needed to move to the more dominant naming convention of `(TargetType)Adapter`

interop/src/main/java/zipkin/interop/CloseAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+package zipkin.interop;
+
+import java.io.Closeable;
+import java.io.IOException;
+import zipkin.async.AsyncSpanStore;
+
+final class CloseAdapter {
+
+  public static void closeQuietly(AsyncSpanStore spanStore) {
+    if (spanStore instanceof Closeable) {
```
anuraaga: Closeable subclasses AutoCloseable so I think it's fine to leave this if out.

interop/src/main/java/zipkin/interop/CallbackWithPromise.java
```diff
@@ -0,0 +1,33 @@
...
+  final Promise<P> promise = Promise$.MODULE$.apply();
+
+  protected abstract P apply(C input);
```
anuraaga: Judging from some other Scala code I've seen, using the not very useful apply may be an idiom, but if not I think convertToScala is more readable ;)

interop/src/main/java/zipkin/interop/ScalaSpanStoreAdapter.java
```diff
@@ -18,36 +18,48 @@
...
 
-  public ScalaSpanStoreAdapter(SpanStore spanStore) {
+  public ScalaSpanStoreAdapter(AsyncSpanStore spanStore) {
     this.spanStore = spanStore;
   }
 
   @Override
+  public Future<BoxedUnit> apply(Seq<Span> input) {
+    VoidCallback callback = new VoidCallback();
+    spanStore.accept(ScalaSpanStoreAdapter.invert(input), callback);
```
anuraaga: Just invert(input) (unless it's specifically zipkin's style to prefix static methods in the same class, it's not totally unreasonable IMO).

zipkin-spanstores/cassandra/pom.xml
```diff
@@ -58,5 +58,15 @@
...
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.zipkin.java</groupId>
+      <artifactId>spanstore-guava</artifactId>
+      <version>0.9.4-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>io.zipkin.java</groupId>
+      <artifactId>spanstore-guava</artifactId>
```
anuraaga: Dupe

interop/src/main/java/zipkin/interop/CallbackWithPromise.java
eirslett: `apply` is basically a substitute for functions, because the JVM doesn't allow functions to exist independent of some class they have to belong to (And then, they're called methods). So `Xxx.apply(yyy)` would probably just be `xxx(yyy)` in a functional language. The Scala compiler actually lets you do that, implemented as syntax sugar - skipping the `apply` part:

```scala
val myCallback = new SomeCallbackWithPromise()
myCallback(input) // this will compile to myCallback.apply(input)
```

If CallbackWithPromise is supposed to be used from Scala, then I think `apply` is a good idea. If it's only Java code, then it might confuse Java programmers who don't know Scala.

anuraaga: It did exactly confuse a Java programmer who doesn't know Scala ;) I did have a feeling this is the reason I see so many apply's in scala+java code though, and don't have anything against it if it's idiomatic to that world (one I personally hope not to venture to deep into ;) ).

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanStore.java
```diff
@@ -30,44 +33,50 @@
...
 
 /**
  * CQL3 implementation of a span store.
  *
  * <p>This uses zipkin-cassandra-core which packages "/cassandra-schema-cql3.txt"
  */
-public final class CassandraSpanStore implements SpanStore, AutoCloseable {
-  static final ThriftCodec THRIFT_CODEC = new ThriftCodec();
-  static final Comparator<List<Span>> TRACE_DESCENDING = new Comparator<List<Span>>() {
+public final class CassandraSpanStore extends AsyncSpanStoreAdapter
```
anuraaga: Since we implemented the elasticsearch store with guava futures in the end I guess we'll want to use the same pattern there too? Doesn't have to be this PR of course

zipkin/src/main/java/zipkin/async/AsyncSpanStoreAdapter.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.concurrent.Executor;
+import zipkin.DependencyLink;
+import zipkin.QueryRequest;
+import zipkin.Span;
+import zipkin.SpanStore;
+import zipkin.internal.Nullable;
+
+import static zipkin.internal.Util.checkNotNull;
+
+public final class AsyncSpanStoreAdapter implements AsyncSpanStore {
```
anuraaga: As I was reviewing this PR my mind was going crazy and couldn't figure out why - finally found out it's because there are two of these same-named classes which I was combining in my head! 

Admittedly I don't have great suggestions but do think we need to fix the names. The obvious ones might be
GuavaToAsyncSpanStoreAdapter + BlockingToAsyncSpanStoreAdapter

but they do seem a bit gross. Another way to think might be that Adapter's in Java tend to be abstract classes to help implement an interface, so the one that converts guava to AsyncSpanStore is truly an adapter, this one is not. But don't have a great suggestion :(

Naming could become less of an issue by hiding into inner static classes and exposing semantically named methods. For example
```
public final class SpanStoreAdapters {
  static AsyncSpanStore blockingToAsync(SpanStore spanStore, Executor executor) {
    return new InternalBlockingToAsyncSpanStoreShim(spanStore, executor);
}
```

Yes, the class name is ridiculous, but it's not public :)

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanStore.java
adriancole: >
> Since we implemented the elasticsearch store with guava futures in the end
> I guess we'll want to use the same pattern there too? Doesn't have to be
> this PR of course
>
it isn't 100% guava futures, but yeah.. it could be without much work. good
point.

anuraaga:
In general, I like the approach :) My biggest concern is some confusing class names - the class names don't seem to say much about what the classes do, which is probably exemplified best by there being two classes with the same name ;) I suggested potentially using static factory methods instead of class constructors, which can provide for more expressive English, hopefully these suggestions can help.

zipkin/src/main/java/zipkin/async/AsyncSpanStoreAdapter.java
adriancole: >
> Admittedly I don't have great suggestions but do think we need to fix the
> names. The obvious ones might be
> GuavaToAsyncSpanStoreAdapter + BlockingToAsyncSpanStoreAdapter
>
thanks for suggesting names.. it has been bothering me, too. I actually had
a XToYAdapter somewhere in history, so will try it on for size :)

interop/src/main/java/zipkin/interop/CloseAdapter.java
adriancole: done

interop/src/main/java/zipkin/interop/CallbackWithPromise.java
adriancole: I agree with @eirslett that apply is a thing in scala, but this code is just an interop test adapter right now.

If we decide to move this adapter into the scala project, which I think I might suggest once we are 1.0, I'd rewrite it in scala and obviate most of the weird lines in these files.

For now, I'll put convertToScala()

interop/src/main/java/zipkin/interop/ScalaSpanStoreAdapter.java
adriancole: done

zipkin-spanstores/cassandra/pom.xml
adriancole: done

zipkin-spanstores/cassandra/src/main/java/zipkin/cassandra/CassandraSpanStore.java
adriancole: done. wasn't much work

zipkin/src/main/java/zipkin/async/AsyncSpanStoreAdapter.java
adriancole: done (renamed). due to some things like field setting, and the base adapter type, I couldn't hide the classes themselves.

anuraaga:
Need to repush?

adriancole:
just repushed!

adriancole:
also rewrote summary

adriancole:
oops forgot to push changes to interop.

anuraaga:
LGTM

PR #128, opened 2016-03-25
Fix all service names and span names not being returned because the d…

anuraaga:
…efault limit of aggregations was 10, not 0.

### DIFF
--- a/zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
+++ b/zipkin-spanstores/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchSpanStore.java
@@ -263,11 +263,13 @@ public class ElasticsearchSpanStore extends GuavaToAsyncSpanStoreAdapter
             .setQuery(matchAllQuery())
             .setSize(0)
             .addAggregation(AggregationBuilders.terms("annotationServiceName_agg")
-                .field("annotations.endpoint.serviceName"))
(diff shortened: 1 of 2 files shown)


adriancole:
thanks. looking into why jdbc failed

anuraaga:
Me too - got my docker mysql up ;) For the service names one, it looks like records with a duplicate binary annotation are deduped by the mysql index. It seems to have to do with mysql index deduping

1) span names seems to mark trace_id+span_id as unique. From my understanding this isn't true (and why we merge)?

2) service names mark trace_id, span_id, key, timestamp as unique. This seems more reasonable and the issue is probably that my test always has two binary annotations with two different service names, which is probably sort of illegal. But first I'm trying to figure out in that case why does getAllServiceNames() pass.

anuraaga:
Ah for #2 it's because the key is different in service and yak, so they don't end up deduping.

Still not too sure if it's ok for jdbc to be deduping based on span id but for now I can just use a unique span id for each span

adriancole:
>
> Me too - got my docker mysql up ;) For the service names one, it looks
> like records with a duplicate binary annotation are deduped by the mysql
> index. It seems to have to do with mysql index deduping
>
> 1) span names seems to mark trace_id+span_id as unique. From my
> understanding this isn't true (and why we merge)?
>
opened
https://github.com/openzipkin/zipkin/issues/939#issuecomment-202230112 on
the span name issue as it is pretty hairy

PR #129, opened 2016-03-25
Fix handling of request.limit, which was being applied pre-filtering …

### LINKED ISSUE #127: Elasticsearch request.limit doesn't work properly
After ingesting some millions of spans I've realized that the pipeline aggregation trick didn't work out for limiting traces as the limit is applied before the filtering happens. I guess this isn't so suprising, backtracking out of a filter to do ordering seemed to good to be true.

I'll try replacing it with two queries, one to get filtered traces, and one to order by unfiltered spans with the limit. Trying it out the performance doesn't seem like it'll be a problem. Admittedly RAM usage within zipkin-server will go up if lots of traces match the filtered query, but trying with ~100K matched traces, the queries each finish in a couple dozen milliseconds so maybe it's ok.

anuraaga:
…instead of post-filtering resulting in empty trace result sets. Now, we break from Zipkin API's guarantees of limiting with respect to traces ordered by first span, instead limiting with respect to order of the first filtered span. In real-world usage, this should not cause any problems and makes sure the query is fast (query filters are better optimized than aggregation filters). The result returned to the client is still sorted correctly as we sort ourselves once more after applying post-processing.

Fixes #127


adriancole:
SGTM

PR #132, opened 2016-03-28
Clarifies zipkin-server's endpoints and how configuration works

adriancole:
This highlights that zipkin-server is a Spring Boot application, as that
clarifies which admin endpoints are available. It also makes explicit
that environment variables are actually bound to yaml configuration.

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -1,10 +1,19 @@
 # zipkin-server
-The hosts the Zipkin [Api](http://zipkin.io/zipkin-api/#/) and [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui).
+zipkin-server is a [Spring Boot](http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/) application, packaged as an executable jar. You need JRE 8+ to start zipkin-server.
 
-Span storage and transports are configurable. By default storage is
-in-memory and the http span transport (POST /spans endpoint) is enabled.
+Span storage and transports are configurable. By default, storage is
+in-memory, the http span transport (POST /spans endpoint) is enabled,
+and the server listens on port 9411.
 
-Note that the server requires minimum JRE 8.
+## Endpoints
+
+The following endpoints are defined for Zipkin:
+* / - [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui)
+* /config.json - [Configuration for the UI](#configuration-for-the-ui)
+* /api/v1 - [Api](http://zipkin.io/zipkin-api/#/)
+* /health - Returns 200 status if OK
+
+There are more [built-in endpoints](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html) provided by Spring Boot, such as `/metrics`. To comprehensively list endpoints, `GET /mappings`.
 
 ## Running locally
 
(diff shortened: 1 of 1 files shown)


adriancole:
cc @gena01 @abesto

zipkin-server/README.md
```diff
@@ -1,10 +1,18 @@
...
 
-Note that the server requires minimum JRE 8.
+## Endpoints
+
+The following endpoints are hosted on the zipkin-server:
+* / - [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui)
+* /api/v1 - [Api](http://zipkin.io/zipkin-api/#/) and [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui)
+* /health, /metrics, ... - [Admin](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html)
+
+Tip: Admin endpoint mappings show up in logs under the category `o.s.b.a.e.mvc.EndpointHandlerMappings`
```
shakuzen: The `/mappings` endpoint is probably easier to reference than the logs.

adriancole: done. thanks for the tip on the tip

abesto:
This makes sense to me, even with little Spring (Boot) experience. :shipit:

zipkin-server/README.md
```diff
@@ -1,10 +1,19 @@
...
+Span storage and transports are configurable. By default, storage is
+in-memory, the http span transport (POST /spans endpoint) is enabled,
+and the server listens on port 9411.
 
-Note that the server requires minimum JRE 8.
+## Endpoints
+
+The following endpoints are defined for Zipkin:
+* / - [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui)
+* /config.json - Configuration for the UI
```
gena01: How do I set/change these? What if I am running a docker container?

gena01:
So the line:
 * `STORAGE_TYPE`: SpanStore implementation: one of `mem` or `mysql`

seems wrong, since there's an example of various other types in the same README file.

zipkin-server/README.md
```diff
@@ -1,10 +1,19 @@
...
+* /api/v1 - [Api](http://zipkin.io/zipkin-api/#/) and [UI](https://github.com/openzipkin/zipkin/tree/master/zipkin-ui)
```
gena01: You can say that UI uses the API. Somehow the way it is right now is a bit unclear to me.

zipkin-server/README.md
```diff
@@ -1,10 +1,19 @@
...
+* /health - Returns 200 status if OK
+
+There are more [built-in endpoints](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html) provided by Spring Boot, such as `/metrics`. To comprehensively list endpoints, `GET /mappings`.
```
gena01: I would add `/metrics` to the list since it can be important (I tend to collect metrics and graph stuff) and leave the line about more options being available with a link.

adriancole: reason this is left out is because it isn't defined by zipkin. Ex
zipkin-scala doesn't define it here, and even if it did, it would be a
different format. I did call out /metrics in the sentence, as I agree it is
important to highlight.

hope this makes partial sense.

zipkin-server/README.md
adriancole: typo. fixed

zipkin-server/README.md
adriancole: added a section

zipkin-server/README.md
adriancole: we can consider promoting /metrics to the standard list of endpoints when the scala project is turned down. Until then, I'm leaving this as-is as the endpoint isn't at the same location, nor has similar contents in the scala server. https://github.com/openzipkin/zipkin/issues/1071

adriancole:
ok I think I addressed the comments, and happy to address more post-merge. I'm going to merge soon so as to prevent known-confusion from persisting.

Thanks for the review @gena01 and @abesto!

Issue #135, opened 2016-03-29
Move storage to a component model


adriancole:
Right now, storage is all routed through a single interface, SpanStore. While convenient, it places responsibilities that are sometimes distant together. For example, most methods are for the zipkin v1 query api, but there's also a span consumer method which is used for collection.

This worked well enough, but it is getting complicated especially with async. It will get more difficult when we support things like dependency inserts (needed for c* and Elasticsearch)

I'd like to move away from an uber interface and towards an object graph. Ex

```java
class CassandraStorage {
  private Repository repository;

  GuavaSpanConsumer spanConsumer()

  GuavaDependenciesConsumer dependenciesConsumer()

  GuavaQueryApi queryApi()

}
```

Something like this is easier to work with, particularly as it can hide internal state like Repository. It is more flexible than relying on inheritance for all the things.

adriancole:
#148 broke up inheritance of InMemoryX, and #151 is taking care of Elasticsearch and Cassandra.

Next step is to make, for example ElasticsearchStorage, which has a builder that can hide its connections and internal types. This would be an object graph that returns guava storage types.

adriancole:
raised #157 to implement this.. tested by removing spring special-casing around duplicate beans, and also having cassandra down, starting the server, then bringing it up later

Issue #133, opened 2016-03-28
Return early on POST /spans


eirslett:
Some people might want to skip Kafka entirely, and make their applications send span data directly to zipkin-server over HTTP. (Maybe via a load balancer like haproxy)
In that case, any problems in zipkin-query would lead to slow HTTP responses, which in turn might affect the instrumented application.

What we need is a "spoon feeding" mechanism that lets instrumented applications get rid of their span data as quickly as possible, fire-and-forget. And then the spans will be fed into the storage backend at its maximum rate.

In zipkin-server, we could return HTTP 202 Accepted, and then add the received spans to an in-memory queue, which would be consumed by a SpanConsumer as quickly as possible. If the queue grows in size to more than some limit (e.g. 64 MB) we would start dropping spans.

The main benefit would be that you could run zipkin without kafka *or* scribe, just (multiple instances of) zipkin-server and the storage backend.

adriancole:
Agreed.

This is why we have a span consumer interface (in fact I'd almost prefer
removing span consumption from SpanStore, except that it is convenient).

Ex. the below is an async span consumer used by the http endpoint, so that
we can return early. Ack that it is easy to miss the Async annotation,
which is what does the magic to make it return early.

https://github.com/openzipkin/zipkin-java/blob/master/zipkin-server/src/main/java/zipkin/server/ZipkinSpanWriter.java

adriancole:
actually, we should be using AsyncSpanConsumer for this stuff..

adriancole:
I have work in progress on this. It is definitely something to clarify prior to 1.0

eirslett:
Ah, I see, I missed the magic annotation.

adriancole:
https://github.com/openzipkin/zipkin-java/pull/136 removes the magic annotation, making async storage more explicit.

adriancole:
<img width="1195" alt="screen shot 2016-05-07 at 8 21 19 pm" src="https://cloud.githubusercontent.com/assets/64215/15212832/a7edc96c-1875-11e6-85e6-5dbbc1f71ba5.png">

eirslett:
:fireworks:

PR #136, opened 2016-03-29
Makes AsyncSpanConsumer the authoritative type for writing spans

adriancole:
Before, `accept(List<Span>)` was on `SpanStore`, a blocking interface.
This helped for a while, but it is clear storage commands should be
async, particularly for reasons listed in #133.

This change removes writes from `SpanStore` and re-organizes the code
around `AsyncSpanConsumer`. This is the first refactor on the way
towards a component model #135.


zipkin-spanstores/guava/src/main/java/zipkin/spanstore/guava/GuavaSpanStore.java
```diff
@@ -26,13 +26,7 @@
...
  * @see zipkin.SpanStore
  */
-public interface GuavaSpanStore extends GuavaSpanConsumer {
-
-  /**
-   * Version of {@link zipkin.SpanStore#accept} that returns {@link ListenableFuture}.
-   */
-  @Override
-  ListenableFuture<Void> accept(List<Span> spans);
+public interface GuavaSpanStore extends GuavaAsyncSpanConsumer {
```
adriancole: still extending this until we have a better modularity story

zipkin-server/src/main/java/zipkin/server/ZipkinSpanWriter.java
```diff
@@ -1,41 +0,0 @@
...
-import org.springframework.stereotype.Service;
-import zipkin.Sampler;
-import zipkin.SamplingSpanStoreConsumer;
-import zipkin.Span;
-import zipkin.SpanConsumer;
-import zipkin.SpanStore;
-
-/** Asynchronously writes spans to storage, subject to sampling policy. */
-@Service
-public class ZipkinSpanWriter implements SpanConsumer {
```
adriancole: this was an async type before. It is now obviated by AsyncSpanConsumer, which has more implementation flexibility (ex you can join cassandra's futures instead of wrapping with `@Async`

Issue #108, opened 2016-03-19
zipkin-server modularity


adriancole:
Currently, zipkin-server needs to know up-front which span stores or transports it can support. This was helpful during the early stages of development, as options were limited, dependencies weren't broad and we've never done modularity in zipkin-scala, yet. As noted in #107 we could do a bit more here, taking advantage of some spring-boot packaging options.

## Option 1: Isolate spring config and move zipkin-server's exec jar packaging from JAR to ZIP layout
One option presented by @anuraaga was to change the exec jar packaging from JAR to ZIP. ZIP uses the PropertiesLoader which is identical to JAR with no options, but also gives the option to add jars to the classpath, for example. It allows easily starting up the server with various plugins.

### Impact
To add a new span store, a jar that bundles spring configuration can be added to `LOADER_PATH`  at runtime. This allows the server to startup with various plugins. This is easy to document and understand from a user POV. ex. add this single file and corresponding configuration

### Work needed
We'd still need to package the spring configuration somewhere, and publish them. We can't forget that zipkin-server is not the only consumer of zipkin-java libraries.

#### Optional dependency of the component
For example, the spring configuration could be an optional dependency of the span stores themselves. We'd have to be careful to test that spring is in fact an optional dependency, as not all users of span stores will be running spring, or the same version of spring as zipkin-server is. For example, this might mean running a nested maven invoker. We may still need to publish a separate artifact, if we need an all-jar (for single-file plugin purposes)

#### Buddy component
Spring configuration could be in a separate maven dependency, which packages an all jar and includes tests that exercise the server. Ex. for spanstore-elasticsearch, the buddy could be spanstore-elasticsearch-plugin, or activator. This adds another explicit maven project, but simplifies and isolates the build/test/publish concern. It also can correctly couple the version of spring-boot to zipkin-server.

## Recap of status quo
Currently, zipkin-server has compile dependencies on all possible span stores and transports, but these are optional at runtime. Tricks like below are used to ensure those classes are indeed optional, and not asked for unless they are really needed.

```java
@ConditionalOnProperty(name = "zipkin.store.type", havingValue = "elasticsearch")
@ConditionalOnClass(name = "zipkin.elasticsearch.ElasticsearchSpanStore")
```
### Why are these optional
[...]

PR #139, opened 2016-03-30
Makes JDBC initialization dependent on storage type

### LINKED ISSUE #137: zipkin-java 0.10 /health endpoint shows wrong information?
I am seeing this when I try to hit /health:
```
    "status": "DOWN",
    "error": "org.springframework.jdbc.CannotGetJdbcConnectionException: Could not get JDBC Connection; nested exception is java.sql.SQLNonTransientConnectionException: Could not connect to address=(host=localhost)(port=3306)(type=master) : Connection refused"
```

The thing is I am not using mysql backend, I am using cassandra. Also that would never work inside the Docker container since mysql would not be running on "localhost" unless the networking is set to "host" from default (bridge)

adriancole:
Before, auto-configuration would initialize JDBC even when the storage
type was cassandra. This caused the health check to false negative.

One way attempted was to move the yaml config to inside zipkin as
opposed to spring. However, this still ended up being read by auto-
configuration. The end result was making this use the same properties
approach as other storage services, disabling auto-configuration in the
yaml file.

This helps us move towards better modularity #108 

Fixes #137

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinMySQLProperties.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinMySQLProperties.java
@@ -1,0 +1,83 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.server;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("mysql")
+public class ZipkinMySQLProperties {
+  private String host = "localhost";
+  private int port = 3306;
+  private String username;
+  private String password;
+  private String db = "zipkin";
+  private int maxActive = 10;
+  private boolean useSsl;
+
+  public String getHost() {
+    return host;
+  }
+
+  public void setHost(String host) {
+    this.host = host;
+  }
+
+  public int getPort() {
+    return port;
+  }
+
+  public void setPort(int port) {
+    this.port = port;
+  }
+
+  public String getUsername() {
+    return username;
+  }
+
+  public void setUsername(String username) {
+    this.username = "".equals(username) ? null : username;
+  }
+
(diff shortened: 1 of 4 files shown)


adriancole:
tested..

```bash
# when STORAGE_TYPE=cassandra
$ curl -s 'localhost:9411/health'
{"status":"UP","diskSpace":{"status":"UP","total":498920849408,"free":418598928384,"threshold":10485760}}
# when STORAGE_TYPE=mysql
$ curl -s 'localhost:9411/health'
{"status":"UP","diskSpace":{"status":"UP","total":498920849408,"free":418602573824,"threshold":10485760},"db":{"status":"UP","database":"MySQL","hello":1}}
```

zipkin-server/src/main/java/zipkin/server/ZipkinMySQLProperties.java
```diff
@@ -0,0 +1,83 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.server;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("mysql")
+public class ZipkinMySQLProperties {
```
adriancole: this class represents the input config from zipkin, which will be mapped directly to non-spring pojo config similar to elasticsearch and cassandra in the future.

zipkin-server/src/main/java/zipkin/server/ZipkinServerConfiguration.java
```diff
@@ -141,19 +143,37 @@ public Object postProcessAfterInitialization(Object bean, String beanName) {
...
+          .driverClassName("org.mariadb.jdbc.Driver")
+          .url(url.toString())
+          .username(mysql.getUsername())
+          .password(mysql.getPassword())
+          .build();
+    }
+
     @Bean JDBCSpanStore jdbcSpanStore() {
-      return new JDBCSpanStore(datasource, new Settings().withRenderSchema(false), listener);
+      return new JDBCSpanStore(dataSource(), new Settings().withRenderSchema(false), listener);
```
shakuzen: The `ConditionalOnMissingBean` on the `dataSource()` method makes it seem like the user can provide their own `DataSource`, but then this line explicitly uses the `dataSource()` method. You should be able to auto-wire the `DataSource` as this Bean method's parameter, I believe.

```java
@Bean JDBCSpanStore jdbcSpanStore(DataSource dataSource) {
  return new JDBCSpanStore(dataSource, new Settings().withRenderSchema(false), listener);
}
```

adriancole: great catch!

adriancole: done

PR #141, opened 2016-03-30
Attempts to de-flake Kafka transport tests

adriancole:
Kafka tests were failing on master. This does a few things to help:
* Starts ZK+Kafka in travis as opposed to via test lifecycle
* Don't exclude log4j: this is actually used by Kafka!
* Configure logback
* Don't try to use "localhost", which can fail. Use 127.0.0.1
* Close more resources
* Sets a hard timeout of 10 seconds to run all Kafka tests
* Ignores `skipsOnConsumerException` test for now, as breaks travis

### DIFF
--- a/zipkin-transports/kafka/pom.xml
+++ b/zipkin-transports/kafka/pom.xml
@@ -14,7 +14,9 @@
     the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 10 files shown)


zipkin-transports/kafka/pom.xml
```diff
@@ -45,24 +43,27 @@
...
-      <!-- excludes eagerly bound logger dependencies,
-           so that they don't leak into zipkin-server -->
+      <!-- don't eagerly bind slf4j -->
       <exclusions>
         <exclusion>
           <groupId>org.slf4j</groupId>
           <artifactId>slf4j-log4j12</artifactId>
         </exclusion>
-        <exclusion>
-          <groupId>log4j</groupId>
```
adriancole: kafka will crash without log4j, depending on how it is initialized.

PR #142, opened 2016-03-30
Fixes small configuration glitches

adriancole:
Fixes self-tracing, removes the storage type registry, and replaces
"zipkin.store.type" with "zipkin.storage.type" which closer relates to
the env variable "STORAGE_TYPE"


adriancole:
this is progress towards #108 as it removes the span store type enum

PR #144, opened 2016-03-30
Applied merging for in memory span store

marcingrzejszczak:
I had some issues in Sleuth with the dependency graph, the graph wasn't displayed properly after clicking around in the application. 

HOWEVER, if I took the spans from Zipkin (/trace/...) and POSTed them back to the app the graph was working fine.

I checked the difference between the spans returned by Zipkin and those that it operates on while trying to build the graph and I saw that this was missing:

` Collection<Span> mergedSpans = CorrectForClockSkew.apply(MergeById.apply(trace));`

Now the graph looks fine. I hope that I'm not applying a hack or sth like that ;)

### DIFF
--- a/zipkin/src/main/java/zipkin/InMemorySpanStore.java
+++ b/zipkin/src/main/java/zipkin/InMemorySpanStore.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+
 import zipkin.async.AsyncSpanConsumer;
 import zipkin.async.AsyncSpanStore;
 import zipkin.async.Callback;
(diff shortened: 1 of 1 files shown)


adriancole:
Clock skew is probably irrelevant for counting spans, but merging indeed sounds important. I will try to make a breaking test for this!

adriancole:
I think this is a general problem, but I don't want to merge until I can reproduce it (I think I have), and fix it for all the things. eta todayish

adriancole:
actually the only stores that this affects is mysql and mem, as they are the only ones that do linking at query time (at the moment)

adriancole:
fwiw mysql seems to be ok bc DependencyLinkSpanIterator merges internally.. just need to make a test to make sure both work

PR #151, opened 2016-04-01
Exposes StorageAdapters and GuavaStorageAdapters, hides incidental types

adriancole:
This prefers transformation to inheritance, which makes wiring more
transparent and reduces the amount of types we need to expose as public.

Ex.
```java
@Bean AsyncSpanConsumer spanConsumer(JDBCSpanStore jdbc, Sampler sampler) {
  AsyncSpanConsumer consumer = blockingToAsync(jdbc::accept, executor());
  return makeSampled(sampler, consumer);
}
```

This introduces some temporary tech debt which should be cleared shortly
when we implement storage components.

See https://github.com/openzipkin/zipkin-java/pull/124#discussion_r57331260


anuraaga:
Skimmed through, LGTM

PR #153, opened 2016-04-02
Renames storage artifacts from spanstore-X to storage-X

adriancole:
The naming convention spanstore-X will not hold up long, particularly
as we already have multiple storage interfaces, and aggregations are
dependency links aren't spans.

This uses the convention `spanstore-x` as it relates better to what
the components do. It also is closer the the environment variable
`STORAGE_TYPE`. ex `STORAGE_TYPE=cassandra` needs `storage-cassandra`.

Also, in #135 we'll end up with storage component classes, like
`ElasticsearchStorage`, which provide a graph to the various storage
interfaces like `SpanStore`.


anuraaga:
LGTM

Issue #155, opened 2016-04-03
Release via git tag


abesto:
Similarly to https://github.com/openzipkin/zipkin/blob/master/RELEASE.md and https://raw.githubusercontent.com/openzipkin/zipkin/master/travis/flow.png, but maybe simpler.

PR #160, opened 2016-04-10
Maven central sync

abesto:
This is both nice in itself, and opens the way for #155


abesto:
Hm, in its current state will the central sync run for SNAPSHOT releases on Travis? (Can investigate, but if someone knows this by heart, that's easier :))

adriancole:
this branch looks to have code to build the plugin, but none of the poms
are setup to run it, so I think it is safe to merge if that's the concern.

adriancole:
build was failing due to a missing travis statement. basically travis was using an old version of maven instead of using the wrapper. fingers crossed!

adriancole:
I've excluded this from being published, since it is internal use. We can always change that later.

PR #163, opened 2016-04-16
First attempt to implement release via Travis CI

abesto:
@adriancole What do you think? I didn't do a full integration test, but many of the individual pieces seem to work.

(Aborted the Travis runs to not mess up anything before a review is done)

### DIFF
--- a/.settings.xml
+++ b/.settings.xml
@@ -29,6 +29,11 @@
       <username>${env.BINTRAY_USER}</username>
       <password>${env.BINTRAY_KEY}</password>
     </server>
+    <server>
+      <id>github.com</id>
+      <username>${env.GH_USER}</username>
+      <password>${env.GH_TOKEN}</password>
+    </server>
   </servers>
 </settings>
 
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+#
+# Copyright 2015-2016 The OpenZipkin Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+#
+
+set -ex
+
+build_started_by_tag() {
+  if [ "${TRAVIS_TAG}" == "" ]; then
+    echo "[Publishing] This build was not started by a tag, publishing publishing snapshot"
+    return 1
+  else
+    echo "[Publishing] This build was started by the tag ${TRAVIS_TAG}, publishing release"
+    return 0
+  fi
+}
+
+is_pull_request() {
+  if [ "${TRAVIS_PULL_REQUEST}" != "false" ]; then
+    echo "[Not Publishing] This is a Pull Request"
+    return 0
+  else
+    echo "[Publishing] This is not a Pull Request"
+    return 1
+  fi
+}
+
+is_travis_branch_master() {
+  if [ "${TRAVIS_BRANCH}" = master ]; then
+    echo "[Publishing] Travis branch is master"
+    return 0
+  else
+    echo "[Not Publishing] Travis branch is not master"
+    return 1
+  fi
+}
+
+check_travis_branch_equals_travis_tag() {
+  #Weird comparison comparing branch to tag because when you 'git push --tags'
+  #the branch somehow becomes the tag value
+  #github issue: https://github.com/travis-ci/travis-ci/issues/1675
+  if [ "${TRAVIS_BRANCH}" != "${TRAVIS_TAG}" ]; then
+    echo "Travis branch does not equal Travis tag, which it should, bailing out."
+    echo "  github issue: https://github.com/travis-ci/travis-ci/issues/1675"
+    exit 1
+  else
+    echo "[Publishing] Branch (${TRAVIS_BRANCH}) same as Tag (${TRAVIS_TAG})"
+  fi
+}
+
+# We don't currently publish internal modules such as benchmarks and interop modules.
+maven_release_args="--batch-mode -s ./.settings.xml -Prelease -pl -:benchmarks,-:interop,-:centralsync-maven-plugin -nsu"
+
+#----------------------
+# MAIN
+#----------------------
+MYSQL_USER=root ./mvnw install -nsu
+
+is_pull_request && exit
+if build_started_by_tag; then
+  check_travis_branch_equals_travis_tag
+  ./mvnw $maven_release_args release:prepare
+  ./mvnw $maven_release_args release:perform
+elif is_travis_branch_master; then
+  ./mvnw $maven_release_args deploy
+fi
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -89,9 +89,6 @@ check_tag_equals_version_in_pom() {
     fi
 }
 
-# We don't currently publish internal modules such as benchmarks and interop modules.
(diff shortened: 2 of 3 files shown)


.settings.xml
```diff
@@ -29,6 +29,11 @@
       <username>${env.BINTRAY_USER}</username>
       <password>${env.BINTRAY_KEY}</password>
     </server>
+    <server>
+      <id>github.com</id>
```
abesto: This part allows us to push the commit and tag created by `maven release:prepare` to GitHub.

abesto:
On second thought: this doesn't take into consideration what tag we push. It dumbly releases the current version and increments the minor version whenever any tag is pushed. We can take one of several directions:

 * Add extra logic to use the version provided in the git tag
 * Use a tag like `trigger-release-20160416` or similar
 * Figure out another way to trigger the release logic

adriancole:
I'd prefer the logic option to take the version from the tag. Ex 1.0.1
releases 1.0.1 and sets next version to 1.0.1-SNAPSHOT

eirslett:
We could compare the version from gradle and the tag name, and fail the build (or reject the git push) if they don't match?

abesto:
@eirslett I like, that's a good way to rule out human error. Not that we ever make mistakes. Looks like I'll have time to look into it this week.

adriancole:
@kristofa this is the thing that will end up being the tag-driven release thing

abesto:
Alright, so https://github.com/openzipkin/zipkin-java/pull/163/commits/3bb3aa12eb3be43027f4855a8e934271fac809e5 should make sure we can release with a tag, and the pom and git tag versions are in sync. But.

The maven release plugin creates a git tag on the commit where it removes `-SNAPSHOT` from the version. This makes sense to me, but it also means we can't use the same tag to trigger the release.

@adriancole Would you be OK with tagging `release-$VERSION` (similarly to `docker-zipkin`)?

adriancole:
>
> @adriancole <https://github.com/adriancole> Would you be OK with tagging
> release-$VERSION (similarly to docker-zipkin)?
>
fine by me!

abesto:
Alright, I think this should work. It obviously won't, and it obviously still won't work after a review, but a review would still be much appreciated.

travis/publish.sh
```diff
@@ -0,0 +1,114 @@
...
+  else
+    echo "[Not Publishing] Travis branch is not master"
+    return 1
+  fi
+}
+
+check_travis_branch_equals_travis_tag() {
+  #Weird comparison comparing branch to tag because when you 'git push --tags'
+  #the branch somehow becomes the tag value
+  #github issue: https://github.com/travis-ci/travis-ci/issues/1675
```
adriancole: thanks for explaining

travis/publish.sh
```diff
@@ -0,0 +1,114 @@
...
+        exit 0
+    elif [[ ! "$tag" =~ ^release-[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$ ]]; then
+        echo "You must specify a tag of the format 'release-0.0.0' to release this project."
+        echo "The provided tag ${tag} doesn't match that. Aborting."
+        exit 1
+    fi
+}
+
+check_tag_equals_version_in_pom() {
+    snapshot_version_in_pom="$(./mvnw -Dexec.executable='echo' -Dexec.args='${project.version}' --non-recursive exec:exec | grep -Ev '(^\[|Download\w+:)')"
```
adriancole: ran it and looks legit

adriancole: another tail could be.. `| sed -n '/Building/s/.* //gp'`

travis/publish.sh
```diff
@@ -0,0 +1,114 @@
...
+  check_release_tag
+  check_tag_equals_version_in_pom
+fi
+
+MYSQL_USER=root ./mvnw install -nsu
+
+if is_pull_request; then
+  true
+elif build_started_by_tag; then
+  ./mvnw $maven_release_args release:prepare
```
adriancole: I've been doing this.. ` ./mvnw -s ./.settings.xml release:prepare`, which does everything except the maven sync

adriancole: correction.. I think the old travis stuff actually did the deploy.

Anyway, you need to run release:prepare on everything, else the non-published modules won't bump version numbers. Probably makes sense to not share args for this reason

adriancole: looking at the old code.. looks like I've never actually run release:perform, and neither did travis!

adriancole:
looks good. only important thing is around release:prepare, as we need that to affect all modules so that versions bump properly.

travis/publish.sh
abesto: > looks like I've never actually run release:perform, and neither did travis!

Confused. Then how do you currently deploy? It's `release:perform`, non? Just not via Travis, always manually from your shell?

abesto:
Gotcha. I have two implementations for doing `release:prepare` but not `release:perform` on internal modules, do you have a preference?

[inline](https://gist.github.com/abesto/7cc82c0116ba0338100d033c4e5752a9) vs [function](https://gist.github.com/abesto/37fa6e68f4c5b28351019d803c31f24b)

travis/publish.sh
adriancole: I do this: ./mvnw -s ./.settings.xml release:prepare

https://github.com/openzipkin/zipkin-java/blob/master/.travis.yml#L37

Then, wait until the build is finished, and click to sync to maven central

adriancole:
inline looks a bit easier to understand

travis/publish.sh
abesto: Alternately, how'd you feel about

```
snapshot_version_in_pom="$(xpath pom.xml '//project/version/text()' 2>/dev/null)"
```

Using `xpath` from Perls `XML::XPath`?

The trade-off: no more heuristics for making output of Maven go away, vs. encoding some understanding of the `pom.xml` into the build script, instead of using the logic of Maven.

travis/publish.sh
abesto: Heh. This is just copy-paste from the original of @jamescway

abesto:
FTR: On Gitter we talked about `release:perform` vs `deploy` (the current solution). We'll see how `release:perform` works first (a change from the currently working setup), since using a single plugin for releases instead of two working in tandem feels nice and cozy. If it doesn't work, we'll fall back to `deploy` when releasing non-snapshot versions. If that doesn't work either, I'll dig in and more deeply understand what and how these plugins do.

adriancole:
sounds good!

travis/publish.sh
adriancole: one issue with relying on the pom is that it introduces a manual step to
update the pom before you do a release.

that's the task that release:prepare does, and release:prepare not only
saves labor updating pom versions, but also chance for error. Can't we read
from the tag instead like the other projects?

adriancole: in case that's cool with folks, here's the batch-mode docs on that
https://maven.apache.org/maven-release/maven-release-plugin/examples/non-interactive-release.html

abesto: We do use `release:perform` and `--batch-mode` to update the POM. This part is about a pre-run sanity check, verifying that the release tag a human created matches the version in the POM file.

abesto: Oh wait, you mean we shouldn't verify the tags against each other, but instead just use the version provided in the tag? That'd indeed remove the need to manually modify the pom when bumping major or minor (as opposed to sub-minor). This goes against the discussion in and around https://github.com/openzipkin/zipkin-java/pull/163#issuecomment-211005586, let's look at it in a separate iteration.

adriancole:
@eirslett @abesto wrt this:

> We could compare the version from gradle and the tag name, and fail the build (or reject the git push) if they don't match?

This makes busy work that doesn't exist in any other zipkin repo (or any repo I work on non-zipkin, including netflix ones). We should read the tag and trust the person who made it. We don't want to discourage people from doing releases as they have more manual work involved like incrementing a patch or subminor first.

By adopting status quo of other projects including the old zipkin, docker etc, we allow people a one-step option to do a release, regardless of whether a minor version change happened or not.

travis/publish.sh
adriancole: This is terrible from an experience POV, departs from all our other
projects, and is directly the opposite desire of the person who's done most
of the releases here!

PR #167, opened 2016-04-25
Rewrites ZooKeeperSampler to eliminate Scala dependency

adriancole:
This implements the adaptive sampler in Java using Apache Curator. The
expected result is the same functionality with fewer build and runtime
dependencies.

Example:

```java
// The ZooKeeperSampler requires you to supply your own Curator.
client = CuratorFrameworkFactory.builder()
    .connectString(zookeeperConnect)
    .retryPolicy(new RetryOneTime(1))
    .build();
client.start();
client.blockUntilConnected(3, TimeUnit.SECONDS);

// This will sample spans based on the capacity of the storage layer.
// Capacity is spans/minute and set by an admin.
sampler = new ZooKeeperSampler.Builder()
    .id("kafka-" + topic + "@" + hostAndPort)
    .build(client);

// Initialize a collector that will sample incoming spans based on
// a coordinated rate.
collector = new KafkaCollector.Builder()
    .zookeeper(zookeeperConnect)
    .topic(topic)
    .writeTo(storage, sampler);
```

### DIFF
--- a/zipkin-samplers/zookeeper/README.md
+++ b/zipkin-samplers/zookeeper/README.md
@@ -2,10 +2,50 @@
 This is an adaptive sampler which can help prevent a surge in traffic
 from overwhelming the zipkin storage layer.
 
-This works in scenarios where you can coordinate the collection/ingestion
-tier and are running zookeeper.
+This works in scenarios where you can coordinate the collection tier
+with ZooKeeper 3.4.x or ZooKeeper 3.5.x.
 
 `zipkin.sampler.zookeeper.ZooKeeperSampler.Builder` includes defaults
-that will against local Zookeeper.
+that will against a given [Curator](http://curator.apache.org) client.
 
-The sampling implementation is compatible with [zipkin-scala](https://github.com/openzipkin/zipkin/tree/master/zipkin-sampler).
+Example:
+
+```java
+// The ZooKeeperSampler requires you to supply your own Curator.
+client = CuratorFrameworkFactory.builder()
+    .connectString(zookeeperConnect)
+    .retryPolicy(new RetryOneTime(1))
+    .build();
+client.start();
+client.blockUntilConnected(3, TimeUnit.SECONDS);
+
+// This will sample spans based on the capacity of the storage layer.
+// Capacity is spans/minute and set by an admin.
+sampler = new ZooKeeperSampler.Builder()
+    .id("kafka-" + topic + "@" + hostAndPort)
+    .build(client);
+
+// Initialize a collector that will sample incoming spans based on
+// a coordinated rate.
+collector = new KafkaCollector.Builder()
+    .zookeeper(zookeeperConnect)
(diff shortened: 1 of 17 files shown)


zipkin-samplers/zookeeper/src/main/java/zipkin/sampler/zookeeper/ZooKeeperSampler.java
```diff
@@ -87,29 +119,144 @@ public ZooKeeperSampler build() {
...
-        builder.windowSize,
-        builder.sufficientWindowSize,
-        builder.outlierThreshold
-    );
+    groupMember = builder.id;
+    boundary =
+        new AtomicLong((long) (Long.MAX_VALUE * builder.initialRate)); // safe cast as less <= 1
+    spanCount = new AtomicInteger(0);
+    storeRate = new AtomicInteger();
+    curator = closer.register(curator(builder));
```
devinsba: It would be nice to be able to give the Sampler a CuratorFramework that is already configured/started. We already have one in our application frameworks that we are using for our dynamic sampler

adriancole: Thanks for the feedback. Will do

adriancole:
@mosesn @eirslett @devinsba PTAL

zipkin-samplers/zookeeper/README.md
```diff
@@ -2,10 +2,47 @@
...
+
+The following paths are relative to `zipkin.sampler.zookeeper.ZooKeeperSampler.basePath`,
+which defaults to "/zipkin/sampler". 
+
+Path | Owner | Description
+--- | --- | ---
+/targetStoreRate | admin | Admin changes this to no more than 95% capacity of the storage layer in spans/minute.
+/sampleRate | leader | Each member of the sample group's `isSampled(Span)` method will pass based on this value. This is updated no more than {sampler.updateInterval} by the leader.
+/storeRates/{sampler.id} | each sampler | Updated per {sampler.updateInterval} to the amount of spans sampled per minute.
+/election | each sampler | This path is used to determine or elect a new leader.
```
adriancole: This is old behavior, but I wonder if we can't use the /sampleRate path for election? That's what needs to be locked.. Any zookeeper people know if this makes sense?

adriancole:
looking into build fail on `ZooKeeperSamplerTest.sampleRateReadFromZookeeper()`..

below looks like an odd smoking gun, though we are trying to read ZK immediately after writing. I'll add tests around it

```
06:43:53.590 [main-EventThread] WARN  z.sampler.zookeeper.ZooKeeperSampler - sampleRate should be between 0 and 1: was 1.0
```

mosesn:
Thanks for digging into this!  Looks great to me =)

adriancole:
@mosesn thanks for having a look!

PR #168, opened 2016-04-26
Updates to Brave 3.6 and makes self-tracing optional

adriancole:
This updates to latest Brave, and makes it possible to configure
self-tracing.

Self tracing exists to help troubleshoot performance of the zipkin-server.

When Brave dependencies are in the classpath, and `zipkin.self-tracing.enabled=true`,
Zipkin will self-trace calls to the api.

[yaml configuration](zipkin-server/src/main/resources/zipkin-server.yml) binds the following environment variables to spring properties:

Variable | Property | Description
--- | --- | ---
SELF_TRACING_ENABLED | zipkin.self-tracing.enabled | Set to false to disable self-tracing. Defaults to true
SELF_TRACING_FLUSH_INTERVAL | zipkin.self-tracing.flush-interval | Interval in seconds to flush self-tracing data to storage. Defaults to 1

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -22,6 +22,19 @@ To run the server from the currently checked out source, enter the following.
 $ ./mvnw -pl zipkin-server spring-boot:run
 ```
 
+## Self-Tracing
+Self tracing exists to help troubleshoot performance of the zipkin-server.
+
+When Brave dependencies are in the classpath, and `zipkin.self-tracing.enabled=true`,
+Zipkin will self-trace calls to the api.
+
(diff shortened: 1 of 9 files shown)


adriancole:
Before this change, those using docker couldn't disable self-tracing eventhough those making custom builds could (by removing the brave dependency)

adriancole:
build failures are unrelated (due to adaptive sampler) and will be solved by #167

eirslett:
How about calling them `SELF_TRACING_ENABLED` and `SELF_TRACING_FLUSH_INTERVAL`?

adriancole:
had that in my history, so I guess that's 2 votes :P

adriancole:
@eirslett done

eirslett:
:shipit:

PR #169, opened 2016-04-26
Names server side of a transport collector

adriancole:
This renames the server side of a transport collector.
Ex. KafkaTransport -> KafkaCollector

This maintains the name "transport" in artifacts, as they might include
a reporter one day, and the documentation are valid for both sides.

See https://github.com/openzipkin/openzipkin.github.io/issues/23


adriancole:
cc @abesto

PR #171, opened 2016-04-27
Adds ElasticsearchSpanStore.MAX_RAW_SPANS to permit large query results

### LINKED ISSUE #159: elasticsearch store getRawTrace return  only 10 spans
When open span view screen, if the trace has more then 10 spans, it won't work. "getRawTrace" return  only 10 spans. 
The UI looks like  https://github.com/openzipkin/zipkin/issues/1085
Set the reuqest size to large size to make it work.

adriancole:
This adds `ElasticsearchSpanStore.MAX_RAW_SPANS`, which controls the maximum
raw spans returned by any trace query, regardless of trace count.

This is similar to `CassandraStorage.maxTraceCols`, which might warrant
renaming.

Fixes #159

### DIFF
--- a/benchmarks/src/main/java/zipkin/benchmarks/SpanBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin/benchmarks/SpanBenchmarks.java
@@ -49,8 +49,9 @@ public class SpanBenchmarks {
   }
 
   @Benchmark
-  public Span buildMinimalClientSpan() {
-    Endpoint client = Endpoint.create("", 172 << 24 | 17 << 16 | 3);
+  public Span buildClientOnlySpan() {
+    Endpoint client = Endpoint.create("query", 172 << 24 | 17 << 16 | 3);
+    Endpoint mysql = Endpoint.create("mysql", 172 << 24 | 17 << 16 | 4, 3306);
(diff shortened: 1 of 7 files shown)


zipkin-storage/elasticsearch/src/main/java/zipkin/elasticsearch/ElasticsearchStorage.java
```diff
@@ -66,21 +68,32 @@ public Builder index(String index) {
       return this;
     }
 
+    /**
+     * The maximum count of raw spans returned in a trace query. Defaults to 10000
```
anuraaga: If we make this configurable, we will also want to either 

1) Add a check that the argument is less than 10000, the default elasticsearch allowed limit
2) Reflect this value in the index template by setting index.max_result_window
3) Document that users will need to update their index settings manually to allow values larger than 10000.

adriancole: made a constant :P

anuraaga:
LGTM - thanks for fixing

adriancole:
@anuraaga thanks for the review!

PR #176, opened 2016-04-29
Splits instrumentation sampling from collector sampling

adriancole:
Instrumentation samplers are before-the-fact. When a trace id is
provisioned, a sampler is applied to it to see if tracing should take
place or not. The decision is made once and propagated downstream.

Collector samplers are after-the-fact. When parts of a trace are
received from multiple reporters, they all need to be retained or
dropped consistently.

Before, we were using the same sampler for both sides. This is a
problem in two ways, but mostly for low-traffic sites.

* Low-traffic entry-points will see an unstable sample rate, at worst
dropping all traces. For example, the default RNG in java applied to
the boundary sampler would only get within 3% of the target rate after
100k decisions. Exact precision could be prefered, especially since the
overhead is still less than a microsecond/decision.

* High-traffic entry-points had low-sampling overhead, but they had a
chance of sampling exactly the same trace ids. This would cause two
unrelated requests to show up in the same trace.

This corrects the situation by not sharing a sampling algorithm between
the collectors and the instrumentation. It introduces a counting sampler
which has 100% precision in groups of 100, and a boundary sampler, which
is an order of magnitude faster, but probablistic with a measured 3%
error rate after 100K samples.

### DIFF
--- a/benchmarks/src/main/java/zipkin/benchmarks/SamplerBenchmarks.java
+++ b/benchmarks/src/main/java/zipkin/benchmarks/SamplerBenchmarks.java
@@ -31,7 +31,9 @@ import org.openjdk.jmh.runner.Runner;
 import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
-import zipkin.Sampler;
+import zipkin.BoundaryTraceIdSampler;
+import zipkin.CountingTraceIdSampler;
+import zipkin.TraceIdSampler;
 
 /**
  * <p>Zipkin v1 uses before-the-fact sampling. This means that the decision to keep or drop the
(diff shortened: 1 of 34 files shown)


adriancole:
ps here's the uncontended benchmarks of the two samplers. again this is a tradeoff (and someone might be able to make a faster counting one with exact precision, too)

```
SamplerBenchmarks.traceIdSampler_boundary                   -9223372036854775808  thrpt   15   868.309 ±  25.891  ops/us
SamplerBenchmarks.traceIdSampler_boundary                    1234567890987654321  thrpt   15   853.044 ±  28.053  ops/us
SamplerBenchmarks.traceIdSampler_counting                   -9223372036854775808  thrpt   15    22.184 ±   2.957  ops/us
SamplerBenchmarks.traceIdSampler_counting                    1234567890987654321  thrpt   15    22.021 ±   2.635  ops/us
```

adriancole:
@kristofa ps this looks like a reversal, but I like to call it.. learning.. yeah that's the ticket

(in other words, trying to use one golden sampling algorithm was my mistake)

PR #177, opened 2016-04-29
Vendors Cassandra Repository class from openzipkin/zipkin

adriancole:
This vendors the Cassandra Repository class for two reasons:

* Most importantly, the pom is incorrect and includes a scala dependency

This causes classpath conflicts with those using other scala libraries,
such as Kafka compiled against 2.10. The attempt to correct this took
hours and was ultimately unfruitful:

https://github.com/openzipkin/zipkin/pull/1104

* Secondly, we can optimize the code

The Repository class was designed for the "slice query" api in the scala
project. This design isn't relevant here, and hosting it makes it easier
to pull apart.


adriancole:
@dsyer @marcingrzejszczak @Arghanil, sorry about the trouble today. I'll cut this asap

PR #178, opened 2016-04-29
Doubles performance of CountingTraceIdSampler while increasing precision

adriancole:
This doubles the performance and increases the precision of the counting
trace ID sampler by using a bitset. After this, both the boundary and
counting sampler support a precision of 0.0001, or 0.01% sample rate.

### DIFF
--- a/zipkin/src/main/java/zipkin/CountingTraceIdSampler.java
+++ b/zipkin/src/main/java/zipkin/CountingTraceIdSampler.java
@@ -13,6 +13,9 @@
  */
 package zipkin;
 
+import java.util.BitSet;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import static zipkin.internal.Util.checkArgument;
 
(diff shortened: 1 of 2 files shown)


zipkin/src/main/java/zipkin/CountingTraceIdSampler.java
```diff
@@ -34,34 +37,35 @@
...
+    int outOf10000 = (int) (rate * 10000.0f);
+    this.bits = new BitSet(10000);
+    for (int i = 0; i < outOf10000; i++) {
+      this.bits.set(i);
+    }
   }
 
+  /** loops over the pre-canned decisions, resetting to zero when it gets to the end. */
   @Override
   public synchronized boolean isSampled(long traceIdIgnored) {
```
adriancole: synchronized as there's still a race between incrementing and reseting the counter

PR #180, opened 2016-04-29
Makes the counting sampler accurate to 100 decisions and more fair

adriancole:
Previously, the counting sampler biased the first trace ids. For
example, if you had a sample rate of 50%, 5K traces would be sampled,
then 5K would not be. This would result in a poor experience for low
traffic sites. This randomizes the decisions, so that they are more
fair, and restores the granularity to 100 decisions.

### DIFF
--- a/zipkin/src/main/java/zipkin/CountingTraceIdSampler.java
+++ b/zipkin/src/main/java/zipkin/CountingTraceIdSampler.java
@@ -14,7 +14,7 @@
 package zipkin;
 
 import java.util.BitSet;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.Random;
 
 import static zipkin.internal.Util.checkArgument;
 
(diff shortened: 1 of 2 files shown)


adriancole:
benchmarks under 4x thread contention are the same. It is still far slower than the boundary sampler, which is unsurprising as the performance tradeoff is expected.

```
Benchmark                                             (traceId)   Mode  Cnt    Score    Error   Units
SamplerBenchmarks.traceIdSampler_boundary  -9223372036854775808  thrpt   15  887.099 ± 16.474  ops/us
SamplerBenchmarks.traceIdSampler_boundary   1234567890987654321  thrpt   15  873.579 ± 26.905  ops/us
SamplerBenchmarks.traceIdSampler_counting  -9223372036854775808  thrpt   15   20.539 ±  3.205  ops/us
SamplerBenchmarks.traceIdSampler_counting   1234567890987654321  thrpt   15   19.963 ±  2.305  ops/us
```

In summary, `CountingTraceIdSampler` should be a good default for those starting with zipkin, as it can obviate "why are my traces not sampling right?" questions. Those running larger sites or heavier traffic can easily switch to `BoundaryTraceIdSampler`, especially as it also takes a percentage as an input.

PR #191, opened 2016-05-03
Add optional configuration: instrumented site(s)

eirslett:
The Zipkin browser extensions will talk to the Zipkin server and
ask it which sites it is supposed to receive traces from. E.g.
you could have multiple Zipkin installations, one for every
environment:

http://www.mysite.com instrumented by http://prod-zipkin.mysite.com
http://dev.mysite.com instrumented by http://dev-zipkin.mysite.com
http://ci.mysite.com instrumented by http://ci-zipkin.mysite.com

By setting the configuration "instrumented" (a regex expression)
we can help the browser extension dispatch trace URLs to the correct
Zipkin server backend.

E.g. for a Zipkin installation at http://dev-zipkin.mysite.com,
setting "instrumented" to ".*:\/\/dev\.mysite\.com\/.*" would mean
all requests accessing dev.mysite.com/* produce traces that
end up in http://dev-zipkin.mysite.com/traces/XXX. The browser
plugin uses that information.

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -64,6 +64,12 @@ zipkin:
     # Default duration to look back when finding traces or dependency links.
     # Affects the "Start time" element in the UI. 7 days in millis
     default-lookback: ${QUERY_LOOKBACK:604800000}
+    # Which sites this Zipkin UI covers. Regex syntax. (e.g. http:\/\/example.com\/.*)
+    # Multiple sites can be specified, e.g.
+    # - .*example1.com
+    # - .*example2.com
+    # Default is "match all websites"
+    instrumented: ${INSTRUMENTED:.*}
 server:
   port: ${QUERY_PORT:9411}
   compression:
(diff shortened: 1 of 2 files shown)


adriancole:
thx!

zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -64,6 +64,12 @@ zipkin:
     # Default duration to look back when finding traces or dependency links.
     # Affects the "Start time" element in the UI. 7 days in millis
     default-lookback: ${QUERY_LOOKBACK:604800000}
+    # Which sites this Zipkin UI covers. Regex syntax. (e.g. http:\/\/example.com\/.*)
+    # Multiple sites can be specified, e.g.
+    # - .*example1.com
+    # - .*example2.com
+    # Default is "match all websites"
+    instrumented: ${INSTRUMENTED:.*}
```
adriancole: whoops.. I'm going to remove the ENV binding in a separate commit, since we aren't generally doing this for UI props.

PR #192, opened 2016-05-03
Adds cassandra client-side self-tracing

adriancole:
This adds client-side self-tracing, propagation penciled in for later.

screen shot 2016-05-04 at 11 05 21 am

screen shot 2016-05-04 at 11 05 56 am

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinServerConfiguration.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinServerConfiguration.java
@@ -13,6 +13,7 @@
  */
 package zipkin.server;
 
+import com.datastax.driver.core.Session;
 import com.github.kristofa.brave.Brave;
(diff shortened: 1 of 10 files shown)


adriancole:
@michaelsembwever I'm keen merging this as it is still better insight than before. The "get-traces" uses an async chain and I'm not sure I can sort it without a change to brave. @eirslett if you have insight from Finn work on async that could help, lemme know. also cc @kristofa

eirslett:
>  The "get-traces" uses an async chain

Where in the code does it happen?

michaelsembwever:
wow. a lot of work went into this!

looks great.

adriancole:
I lowered the scope of this to just client-side for now, as it will make it more clear what we do when there's an instrumented cluster behind us. I've also included query details in the trace.

I'll open up something else to get to the bottom of the missing spans and copy @eirslett on that.

PR #196, opened 2016-05-05
Fixes self-tracing auto-configuration by quitting ConditionalOnClass

adriancole:
ConditionalOnClass lookups have proven unreliable and extremely
difficult to unwind. For example, the behavior is inconsistent between
the JRE on alpine linux and on a macbook. Rather than continue losing
time on matters like these, this moves to a more explicit condition.


adriancole:
This was hard to test. basically I had to build local docker images, testing with and without the SELF_TRACING_ENABLED flag. We need CI and some tests in the docker repo!

PR #197, opened 2016-05-05
Downgrade interop module to Java 7 to support Eclipse users

dsyer:
There's a bug in Eclipse JDT which makes the interop project
uncompilable with Java 8
(https://bugs.eclipse.org/bugs/show_bug.cgi?id=468276). Since
only one small Java 8 feature was being used in one test
in the whole project, it seems like a small price to pay
to sacrifice that for being able to compile and run in the IDE.

### DIFF
--- a/interop/pom.xml
+++ b/interop/pom.xml
@@ -119,6 +119,13 @@
         </configuration>
       </plugin>
       <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


dsyer:
I have test failures in zipkin-junit as well, but they seem unrelated.

adriancole:
weird.. but sure.

PR #205, opened 2016-05-06
Adds propagation to cassandra self-tracing

adriancole:
This adds propagation to cassandra self-tracing. When the receiver has
zipkin instrumentation installed, zipkin self-traces will propagate and
include the underlying storage operations

screen shot 2016-05-06 at 10 07 14 pm


adriancole:
FYI @eirslett @yurishkuro @danchia @michaelsembwever @kristofa 

This uses finagle's binary encoding for trace ids recently added to Brave in https://github.com/openzipkin/brave/pull/168

Once we have the cassandra tracing finalized, we can add it to our docker image so folks can play with it without compiling this: https://github.com/thelastpickle/cassandra-zipkin-tracing/tree/mck/brave-3.5

PR #207, opened 2016-05-06
Improves json codec performance by being more efficient with hex codec

adriancole:
Before, we were very slow at hex encoding, and this made the json
encoding notably slower than necessary. We were also inefficient by
using regex to check a hex string. This fixes both issues, notably
improving encoding performance.

```
CodecBenchmarks.readLocalSpan_json_java  thrpt   15  155.571 ± 10.273  ops/ms
CodecBenchmarks.writeLocalSpan_json_java  thrpt   15  131.598 ± 11.653  ops/ms

CodecBenchmarks.readLocalSpan_json_java  thrpt   15  162.043 ± 2.652  ops/ms
CodecBenchmarks.writeLocalSpan_json_java  thrpt   15  203.380 ± 6.892  ops/ms
```

### DIFF
--- a/zipkin/src/test/java/zipkin/internal/UtilTest.java
+++ b/zipkin/src/test/java/zipkin/internal/UtilTest.java
@@ -27,6 +27,7 @@ import static org.junit.Assert.assertTrue;
 import static zipkin.internal.Util.equal;
 import static zipkin.internal.Util.lowerHexToUnsignedLong;
 import static zipkin.internal.Util.midnightUTC;
+import static zipkin.internal.Util.toLowerHex;
 
 public class UtilTest {
   @Test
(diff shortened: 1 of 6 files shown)


zipkin/src/test/java/zipkin/internal/UtilTest.java
```diff
@@ -79,4 +80,14 @@ public void lowerHexToUnsignedLongTest() throws ParseException {
 
     }
   }
+
```
eirslett: Maybe also test some of the middle values? Like cafebabe = 3405691582 ?

adriancole: done! b9e4315

PR #213, opened 2016-05-07
The release plugin needs to be told to skip tests differently

abesto:
By the time we do release:prepare and release:perform, tests have
already run and passed.

http://stackoverflow.com/questions/8685100/how-can-i-get-maven-release-plugin-to-skip-my-tests


adriancole:
thanks!

PR #214, opened 2016-05-07
Document release process

abesto:
I'm also planning to provide graphs similar to the ones shown in https://github.com/openzipkin/zipkin/blob/master/RELEASE.md. They'll be considerably simpler here.

Just realized when looking at the release process of `zipkin`: the automation in `zipkin-java` currently has no support for releasing RCs. May not be a problem, and it keeps the flow simpler.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,25 @@
+# Zipkin Release Process
+
+This repo uses semantic versions with a twist: we update minors on api-breaks until we hit 1.0. Please keep this
+in mind when choosing version numbers.
+
+1. **Alert others you are releasing**
+
+   There should be no commits made to master while the release is in progress (about 10 minutes). Before you start
+   a release, alert others on [gitter](https://gitter.im/openzipkin/zipkin) so that they don't accidentally merge
+   anything. If they do, and the build fails because of that, you'll have to recreate the release tag described below.
+
+1. **Push a git tag**
+
+   The tag should be of the format `release-N.M.L`, for example `release-0.18.1`. Note that the release automation
+   will verify that the version in the tag matches the version in `pom.xml` (minus `-SNAPSHOT`), and abort the release
+   if it's not.
+
+1. **Wait for Travis CI**
+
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,22 @@
+# Zipkin Release Process
+
+1. **Push a git tag**
+
+   The tag should be of the format `release-N.M.L`, for example `release-0.18.1`. Note that the release automation
+   will verify that the version in the tag matches the version in `pom.xml` (minus `-SNAPSHOT`), and abort the release
```
adriancole: this is a regression.

This should be fixed ASAP as we now have the same step count as before, except we don't have instructions for the step of applying the new version!

adriancole: I added an issue on this, which summarizes my various protests on the topic https://github.com/openzipkin/zipkin-java/issues/215

adriancole:
added docs on how to increment the version until we fix that step in #215

adriancole:
@abesto regardless of the point about minor releases, this is great work. I've very happy that we've a way to automate releases, especially that patch releases are simpler now.

adriancole:
trying this now for release-0.18.2

adriancole:
last release failed at 409 at the end syncing to maven.. not sure if that's a real 409 or an unauthorized in disguise..

adriancole:
trying again as I don't think the 409 thing is new https://travis-ci.org/openzipkin/zipkin-java/builds/129086407

adriancole:
woot all green

adriancole:
oops looked at wrong repo.. it blipped because a commit came in while it was working..

adriancole:
turns out we can't commit to master while the release thing is in-flight. Updated docs.

adriancole:
ok release things work now, so this is valid and/or ready for polishing

adriancole:
ok. really-really works now. key insight is that the version commit on master is the job that deploys and syncs to maven central

adriancole:
okie ready to merge (all processes are stable now)

PR #216, opened 2016-05-08
Fix #215: directly use version from release tag

### LINKED ISSUE #215: Fix regression where making a release requires snapshot version to match release version
In openzipkin/zipkin, openzipkin/zipkin-java etc we do not rely on a version match to apply the next version. This means that when you tag a release, that's the version it gets. Maven's release plugin doesn't require this either! In fact, if you run the release task, it takes whatever version you say as the version.

I've come to take this behavior for granted the past few years in projects in and outside zipkin. It reduces the busy-work required to make a minor release, which in turn lowers the barrier to release often. This is particularly a problem when using semantic versions on a new project. Ex. zipkin-java has had 18 minor versions, and zipkin itself has so far had 40 in major version 1!

I've strongly voiced against the action taken because I'm the one most impacted. In the last year, I've done hundreds of releases, if you collect the zipkin-scala, the docker repos, the satellites etc. As a releaser in *any* project, I'd protest against intentionally making a minor version update to require 2 manual steps (in this case updating a pom, then doing the release).

If it were the case that there was a technical difficulty, or unimplemented feature, that's cool. For example, [docker-zipkin-java's release process doesn't currently bump versions](https://github.com/openzipkin/docker-zipkin-java/issues/1), but it could. The difference here is that the release process here, we intentionally made it 2 steps!

Let's fix it.

See https://github.com/openzipkin/zipkin-java/pull/163#issuecomment-211005586 to add policy to check that snapshot

abesto:


adriancole:
oops!

./travis/publish.sh: line 109: unexpected EOF while looking for matching `"'

adriancole:
re-pushed, matching the "

abesto:
Oops, thanks for that. Anything else left to do before merging this?

adriancole:
nope was just waiting for green. thanks!

Issue #187, opened 2016-05-01
Remove I/O from all component constructors


adriancole:
Currently, wiring up zipkin is complex because people have to dodge transient connection failures during wiring. Rather than try to expose a `Lazy` type, we could instead just make the implementations not perform I/O in their constructors. In other words, any laziness is an implementation detail.

When this is the case, it will be particularly easier to make a collector. Right now, we have to play games to defer access to the storage component.

adriancole:
Chatted with @anuraaga and seems we can do this as long as we do smarter health checks #140

PR #217, opened 2016-05-10
Fixes crash where there's a race between storage and query components

adriancole:
Formerly, wiring `ZipkinQueryApiV1` would crash the app if a storage
connection was down. In other words, it was doing I/O in the
constructor. The storage component is lazy, so the way to fix this is to
defer access until it is needed.

This also adds a test so that it is found before docker packaging.

See #187


zipkin-server/src/it/minimal-dependencies/src/main/java/zipkin/minimal/ZipkinServer.java
```diff
@@ -1,28 +0,0 @@
...
- */
-package zipkin.minimal;
-
-import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.boot.builder.SpringApplicationBuilder;
-import zipkin.server.EnableZipkinServer;
-
-@SpringBootApplication
-@EnableZipkinServer
-public class ZipkinServer {
```
adriancole: removed this as we should be testing that the real zipkinserver class has no required dependency on optional classes. This class got in the way of that.

zipkin-server/src/it/minimal-dependencies/src/test/java/zipkin/minimal/ZipkinServerTest.java
```diff
@@ -39,7 +40,7 @@
 @SpringApplicationConfiguration(classes = ZipkinServer.class)
 @RunWith(SpringJUnit4ClassRunner.class)
 @WebAppConfiguration
-@TestPropertySource(properties = {"zipkin.store.type=mem", "spring.config.name=zipkin-server"})
```
adriancole: drift..

zipkin-server/src/main/java/zipkin/server/ZipkinQueryApiV1.java
```diff
@@ -49,32 +47,30 @@
   @Value("${zipkin.query.lookback:86400000}")
   int defaultLookback = 86400000; // 7 days in millis
 
-  private final SpanStore spanStore;
+  private final StorageComponent storage;
   private final Codec jsonCodec;
 
-  /** lazy so transient storage errors don't crash bootstrap */
```
adriancole: this didn't solve the bootstrap problem, it just moved it a few seconds later.

PR #222, opened 2016-05-12
Enables maven sync following a successful version tag deployment

### LINKED ISSUE #131: Automatic sync to maven central
Since we already upload releases to bintray, it appears straightforward to also have the releases copied over to central:

https://bintray.com/docs/usermanual/uploads/uploads_syncingartifactswithmavencentral.html

adriancole:
Fixes #131

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -115,6 +115,10 @@ if is_pull_request; then
 #   - If a version tag fails to deploy for a transient reason, delete the broken version from bintray and click rebuild
 elif is_travis_branch_master || is_version_tag; then
   ./mvnw --batch-mode -s ./.settings.xml -Prelease -nsu -pl -:benchmarks,-:interop,-:centralsync-maven-plugin -DskipTests deploy
+
+  # If the deployment succeeded, sync it to Maven Central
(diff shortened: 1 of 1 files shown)


adriancole:
@abesto this look right to you?

travis/publish.sh
```diff
@@ -115,6 +115,12 @@ if is_pull_request; then
 #   - If a version tag fails to deploy for a transient reason, delete the broken version from bintray and click rebuild
 elif is_travis_branch_master || is_version_tag; then
   ./mvnw --batch-mode -s ./.settings.xml -Prelease -nsu -pl -:benchmarks,-:interop,-:centralsync-maven-plugin -DskipTests deploy
+
+  # If the deployment succeeded, sync it to Maven Central
+  if [[ $? -eq 0 && is_version_tag ]]; then
```
abesto: You can omit the `$?` check since `-e` is in effect, so the script will bail out if any of the commands exit with non-zero exit code. Then then condition can be written like just `if is_version_tag; then ...`. Some prefer a guard-like syntax, like so:

```
is_version_tag && ./mvnw ...
```

abesto:
One minor comment, otherwise LGTM.

travis/publish.sh
adriancole: good idea!

adriancole:
ok doing a token-release to test it

abesto:
Oh, uh... One thing I forgot. This guard-like syntax doesn't really play nice with `-e`; the condition exits with non-zero, so `-e` forces the whole script to terminate. Sorry about that.

PR #224, opened 2016-05-12
Deploys snapshots and versions from master

### LINKED ISSUE #220: double-mint deploys
there's 409 on a master build job here: https://travis-ci.org/openzipkin/zipkin-java/builds/129131191
because it was already released on the release tag here: https://travis-ci.org/openzipkin/zipkin-java/builds/129128621

I'm guessing one of these should run, but not both

adriancole:
Since our release task makes tags, not branches, version commits end up
on master. These deploy, so we can't also build the tag, too.

This consolidates code around deployment to the master branch, and uses
the maven version to tell if it should be synchronized to maven central
or not.

Fixes #220
Partially reverts a4bf8e1c47eae314c44fac075c89797817c93876

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,12 @@ install:
 script:
   - ./travis/publish.sh
 
+# Don't build release tags. This avoids publish conflicts because the version commit exists both on master and the release tag.
+# See https://github.com/travis-ci/travis-ci/issues/1532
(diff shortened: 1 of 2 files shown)


adriancole:
@abesto knock-on-wood this is the last mole to whack

PR #227, opened 2016-05-13
Extracts Component, which has a check() method for determining health

### LINKED ISSUE #140: Smarter health checks
zipkin storage are written to not have a hard-dependency on spring. Sometimes they use components, like DataSource, which spring has built-in health checks for. Other times, they use libraries that aren't the same version as spring products, for example cassandra's Repository type isn't the same as Spring Data Cassandra.

We can address this in a number of ways, but I've found it helpful to defer health check to storage component  itself (#135). For example, the designer of elasticsearch can come up with a health check that's more valid for zipkin. Failing that, we can revert to a "cheap" operation, such as listing services.

adriancole:
This pulls out an interface: Component, which has the ability to report
its health via a `check()` method. This is ported to all storage and
collector components. Notably, this also fixes a crash when Kafka isn't
running when zipkin was started.

Fixes #140

Inspired by:
* https://github.com/Netflix/denominator/blob/master/core/src/main/java/denominator/CheckConnection.java
* https://github.com/dropwizard/metrics/blob/3.1-maintenance/metrics-healthchecks/src/main/java/com/codahale/metrics/health/HealthCheck.java

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@
     <cassandra-driver-core.version>3.0.1</cassandra-driver-core.version>
     <moshi.version>1.1.0</moshi.version>
     <okio.version>1.8.0</okio.version>
+    <jooq.version>3.8.0</jooq.version>
     <spring-boot.version>1.3.5.RELEASE</spring-boot.version>
     <!-- MySQL connector is GPL, even if it has an OSS exception.
          https://www.mysql.com/about/legal/licensing/foss-exception/
@@ -213,6 +214,12 @@
         <artifactId>cassandra-driver-core</artifactId>
         <version>${cassandra-driver-core.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.jooq</groupId>
+        <artifactId>jooq</artifactId>
+        <version>${jooq.version}</version>
+      </dependency>
+      <!-- End spring-boot-dependencies overrides -->
 
       <dependency>
         <groupId>com.squareup.moshi</groupId>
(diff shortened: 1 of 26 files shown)


adriancole:
Tested by making various components fail.

For example, starting with a wrong zookeeper like this..

```bash
KAFKA_ZOOKEEPER=idontexist java -jar ./zipkin-server/target/zipkin-server-*exec.jar
```

It starts running, but puts a failed health check (503 code)
```bash
$ curl -s localhost:9411/health|jq .
{
  "status": "DOWN",
  "zipkinKafkaCollector": {
    "status": "DOWN",
    "error": "org.I0Itec.zkclient.exception.ZkException: Unable to connect to idontexist"
  },
  "zipkinScribeCollector": {
    "status": "UP"
  },
  "zipkinStorage": {
    "status": "UP"
  },
  "diskSpace": {
    "status": "UP",
    "total": 498920849408,
    "free": 413605081088,
    "threshold": 10485760
  }
}
```

pom.xml
```diff
@@ -52,6 +52,7 @@
     <cassandra-driver-core.version>3.0.1</cassandra-driver-core.version>
     <moshi.version>1.1.0</moshi.version>
     <okio.version>1.8.0</okio.version>
+    <jooq.version>3.8.0</jooq.version>
```
adriancole: health check found a classpath conflict!

zipkin-server/src/main/java/zipkin/server/brave/TracedSession.java
```diff
@@ -102,7 +103,12 @@ public static Session create(Session target, Brave brave, SpanCollector collecto
       brave.clientSpanThreadBinder().setCurrentSpan(null);
       return new BraveResultSetFuture(target.executeAsync(statement), brave);
     }
-    return method.invoke(target, args);
+    try {
+      return method.invoke(target, args);
+    } catch (InvocationTargetException e) {
+      if (e.getCause() instanceof RuntimeException) throw e.getCause();
```
adriancole: another bug found with the health check

PR #230, opened 2016-05-14
Adds cache headers to /api/v1/services and /api/v1/spans

### LINKED ISSUE #229: Add cache headers to /api/v1/services and /api/v1/spans
In the old codebase, we had logic like the following to help from pummeling the api server when people click around in the UI

```
if (services < 3) add "Cache-Control: max-age=300, must-revalidate"
```

This is because the UI calls the /api/v1/services quite a lot, but the actual values don't change often. The extra guard was to ensure those getting started are able to see traces coming in.

Besides /api/v1/services, /api/v1/spans is also called a lot.

If we add this behavior to zipkin-server, we should probably do the same for spans, as they are quite related.

adriancole:
Same as https://github.com/openzipkin/zipkin/issues/718, except
including the span names endpoint (/api/v1/spans)

Defaults to 300 seconds, overridable via zipkin.query.names-max-age

Fixes #229

### DIFF
--- a/zipkin-server/src/main/java/zipkin/server/ZipkinQueryApiV1.java
+++ b/zipkin-server/src/main/java/zipkin/server/ZipkinQueryApiV1.java
@@ -14,9 +14,12 @@
 package zipkin.server;
 
 import java.util.List;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


mansu:
r+

PR #234, opened 2016-05-16
Hides zipkin-ui version complexity from end users

adriancole:
Currently, users have to track 2 zipkin versions to function:
* 0.X - for the almost all of the code
* 1.X - for the UI

This double-maintenance is a reality for zipkin developers, until we
complete the transition to a single repository. However, we shouldn't
subject users to that, particularly as they don't know when they should
update the UI version (for example that version changes if unrelated
scala code is released).

By repackaging zipkin-ui here, we can hide this complexity from users,
and better prepare ourselves to transition to building the UI using
https://github.com/eirslett/frontend-maven-plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,6 +24,7 @@
 
   <modules>
     <module>zipkin</module>
+    <module>zipkin-ui</module>
     <module>zipkin-junit</module>
     <module>zipkin-guava</module>
     <module>benchmarks</module>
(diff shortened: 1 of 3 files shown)


adriancole:
cc @eirslett @dsyer @marcingrzejszczak

eirslett:
Lgtm!

PR #235, opened 2016-05-16
Namespaces spring configuration to match package names

adriancole:
Before, spring configuration keys for storage or collectors were top-
level. For example, `kafka.X` or `cassandra.Y`. This clutters up the
namespace and interferes with modularity. These properties are now
namespaced like `zipkin.collector.kafka.X` or
`zipkin.storage.cassandra.Y`.

See #108

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -1,46 +1,3 @@
-cassandra:
-  # Comma separated list of hosts / ip addresses part of Cassandra cluster.
-  contact-points: ${CASSANDRA_CONTACT_POINTS:localhost}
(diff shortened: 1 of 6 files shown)


zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -50,13 +7,56 @@ zipkin:
...
   query:
     # 7 days in millis
     lookback: ${QUERY_LOOKBACK:86400000}
     # The Cache-Control max-age (seconds) for /api/v1/services and /api/v1/spans
     names-max-age: 300
   storage:
     type: ${STORAGE_TYPE:mem}
+    cassandra:
+      # Comma separated list of hosts / ip addresses part of Cassandra cluster.
+      contact-points: ${CASSANDRA_CONTACT_POINTS:localhost}
```
adriancole: note: we didn't change the env variables, so this should be a transparent change

PR #236, opened 2016-05-17
Extracts auto-configuration modules from zipkin-server

adriancole:
This decouples zipkin-server from ui, storage and collector modules. 
While there remains an optional dependency, this is only a mechanism to 
include these modules in the fat jar.

To do this, it uses the AutoConfigure technique from Spring Boot,
notably registering a manifest for each component to export.

To completely decouple, some work was needed. Particularly health check
registration needed to become more generic, and occur at the very end of
the application bootstrap so as to not interfere with initialization.

Finally, this clears up a long-standing confusion where we named a 
storage component jdbc, eventhough we only test it with mysql and all 
the flags are named mysql. This settles on mysql as for example 
PostgreSQL would likely need different syntax.

Here's the new composite health check output:

```bash
$ curl -s localhost:9411/health|jq .
{
  "status": "UP",
  "zipkin": {
    "status": "UP",
    "ScribeCollector": {
      "status": "UP"
    },
    "MySQLStorage": {
      "status": "UP"
    }
  },
  "diskSpace": {
    "status": "UP",
    "total": 498920849408,
    "free": 412974030848,
    "threshold": 10485760
  }
}
```


adriancole:
@anuraaga @marcingrzejszczak @eirslett @shakuzen I plan to merge this tomorrow unless there's feedback against.

zipkin-autoconfigure/collector-kafka/src/main/java/zipkin/autoconfigure/collector/kafka/KafkaEnabledCondition.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+/**
+ * This condition passes when Kafka classes are available and {@link
+ * ZipkinKafkaCollectorProperties#getZookeeper()} is set.
+ */
+final class KafkaEnabledCondition extends SpringBootCondition {
```
anuraaga: Do you need this? ZipkinKafkaCollectorAutoConfiguration has a ConditionalOnProperty, and this one doesn't seem to be checking anything else but another property.

anuraaga:
Just one comment, I think I found an experimental class leftover, otherwise LGTM

zipkin-autoconfigure/collector-kafka/src/main/java/zipkin/autoconfigure/collector/kafka/KafkaEnabledCondition.java
marcingrzejszczak: Yeah actually it seems that the Javadoc tells that the class does more than it actually does. Shouldn't you check also for class presence?

eirslett:
LGTM!

zipkin-autoconfigure/collector-kafka/src/main/java/zipkin/autoconfigure/collector/kafka/KafkaEnabledCondition.java
adriancole: The issue is that the normal spring property check doesn't have a way to deal with empty, and the yaml file has a default to empty clause in it. I can update the doc about that.

```
      zookeeper: ${KAFKA_ZOOKEEPER:}
```

@marcingrzejszczak since this is in a module by itself now, do we need a class presence clause anymore? Ex. if you are including this wouldn't you have the classes that it depends on?

zipkin-autoconfigure/collector-kafka/src/main/java/zipkin/autoconfigure/collector/kafka/KafkaEnabledCondition.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package zipkin.autoconfigure.collector.kafka;
+
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+/**
+ * This condition passes when {@link ZipkinKafkaCollectorProperties#getZookeeper()} is set to
```
adriancole: @marcingrzejszczak @anuraaga lemme know if this is a bit better

adriancole:
ps renamed etc the kafkazookeeper condition. It is package private and happy to revisit later, if folks have more suggestions

anuraaga:
In that case, is there any point to the ```zipkin.collector.kafka.enabled``` property? It seems redundant vs setting ```zipkin.collector.kafka.zookeeper``` to empty. On the flip side, we can keep both properties but make the bean ConditionalOnProperty(zipkin.collector.kafka.enabled) while having the presence of zipkin.collector.kafka.zookeeper validated as a precondition somewhere when that bean is processed.

adriancole:
@anuraaga good point. I'll remove the "enabled" property. that was mostly a copy/paste thing, as you're right, we have another means to control enabling of kafka

adriancole:
@anuraaga fixed via d7ce2c4 thx!

PR #238, opened 2016-05-18
Revert "Namespaces spring configuration to match package names"

adriancole:
Config property namespacing broke environment variable mapping.
This reverts that and also fixes a pom glitch in elasticsearch.

### DIFF
--- a/zipkin-server/src/main/resources/zipkin-server.yml
+++ b/zipkin-server/src/main/resources/zipkin-server.yml
@@ -1,3 +1,46 @@
+cassandra:
+  # Comma separated list of hosts / ip addresses part of Cassandra cluster.
+  contact-points: ${CASSANDRA_CONTACT_POINTS:localhost}
+  # Name of the datacenter that will be considered "local" for latency load balancing. When unset, load-balancing is round-robin.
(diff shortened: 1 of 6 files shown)


zipkin-server/src/main/resources/zipkin-server.yml
```diff
@@ -1,3 +1,46 @@
...
+mysql:
+  host: ${MYSQL_HOST:localhost}
+  port: ${MYSQL_TCP_PORT:3306}
+  username: ${MYSQL_USER:}
+  password: ${MYSQL_PASS:}
+  db: ${MYSQL_DB:zipkin}
+  max-active: ${MYSQL_MAX_CONNECTIONS:10}
+  use-ssl: ${MYSQL_USE_SSL:false}
+elasticsearch:
+  cluster: ${ES_CLUSTER:elasticsearch}
```
adriancole: I tested this, btw. it works. IOTW, you can use ES_CLUSTER instead of ELASTICSEARCH_CLUSTER

shakuzen:
Just eyeballing it, I don't see what the problem was. It should have worked how it was.

PR #240, opened 2016-05-19
Updates Cassandra schema to include default ttls and adds upgrade logic

michaelsembwever:
Improvements to the Cassandra storage
 - the schema now defines default TTL for each table
 - deprecates manually controlling span and index ttl
 - supports upgrading the schema

See #240

### DIFF
--- a/zipkin-server/README.md
+++ b/zipkin-server/README.md
@@ -101,10 +101,8 @@ supports version 2.2+ and applies when `STORAGE_TYPE` is set to `cassandra`:
     * `CASSANDRA_CONTACT_POINTS`: Comma separated list of hosts / ip addresses part of Cassandra cluster. Defaults to localhost
     * `CASSANDRA_LOCAL_DC`: Name of the datacenter that will be considered "local" for latency load balancing. When unset, load-balancing is round-robin.
     * `CASSANDRA_MAX_CONNECTIONS`: Max pooled connections per datacenter-local host. Defaults to 8
-    * `CASSANDRA_ENSURE_SCHEMA`: Ensuring that schema exists, if enabled tries to execute script /zipkin-cassandra-core/resources/cassandra-schema-cql3.txt. Defaults to true
+    * `CASSANDRA_ENSURE_SCHEMA`: Ensuring cassandra has the latest schema. If enabled tries to execute scripts in the classpath prefixed with `cassandra-schema-cql3`. Defaults to true
     * `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD`: Cassandra authentication. Will throw an exception on startup if authentication fails. No default
-    * `CASSANDRA_SPAN_TTL`: Time-to-live in seconds for span data. Defaults to 604800 (7 days)
-    * `CASSANDRA_INDEX_TTL`: Time-to-live in seconds for index data. Defaults to 259200 (3 days)
 
 Example usage:
 
--- a/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
+++ b/zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
@@ -15,7 +15,9 @@ package zipkin.storage.cassandra;
 
 import com.datastax.driver.core.BoundStatement;
 import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.RegularStatement;
 import com.datastax.driver.core.Session;
+import com.datastax.driver.core.querybuilder.Insert;
 import com.datastax.driver.core.querybuilder.QueryBuilder;
 import com.datastax.driver.core.utils.Bytes;
 import com.google.common.base.Function;
@@ -28,7 +30,6 @@ import java.nio.charset.CharacterCodingException;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Random;
 import java.util.Set;
 import org.slf4j.Logger;
@@ -55,7 +56,9 @@ final class CassandraSpanConsumer implements GuavaSpanConsumer {
   private final Session session;
   private final TimestampCodec timestampCodec;
   private final int bucketCount;
+  @Deprecated
   private final int spanTtl;
+  @Deprecated
   private final int indexTtl;
   private final PreparedStatement insertSpan;
   private final PreparedStatement insertServiceName;
@@ -64,7 +67,7 @@ final class CassandraSpanConsumer implements GuavaSpanConsumer {
   private final PreparedStatement insertTraceIdBySpanName;
   private final PreparedStatement insertTraceIdByAnnotation;
   private final PreparedStatement insertTraceIdBySpanDuration;
-  private final Map<String, String> metadata;
+  private final Schema.Metadata metadata;
 
   private final ThreadLocal<Set<String>> writtenNames = new ThreadLocal<Set<String>>() {
     private long cacheInterval = toCacheInterval(System.currentTimeMillis());
@@ -97,64 +100,63 @@ final class CassandraSpanConsumer implements GuavaSpanConsumer {
     this.indexTtl = indexTtl;
     this.metadata = Schema.readMetadata(session);
     insertSpan = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("traces")
             .value("trace_id", QueryBuilder.bindMarker("trace_id"))
             .value("ts", QueryBuilder.bindMarker("ts"))
             .value("span_name", QueryBuilder.bindMarker("span_name"))
-            .value("span", QueryBuilder.bindMarker("span"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("span", QueryBuilder.bindMarker("span"))));
 
     insertServiceName = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("service_names")
-            .value("service_name", QueryBuilder.bindMarker("service_name"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("service_name", QueryBuilder.bindMarker("service_name"))));
 
     insertSpanName = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("span_names")
             .value("service_name", QueryBuilder.bindMarker("service_name"))
             .value("bucket", QueryBuilder.bindMarker("bucket"))
-            .value("span_name", QueryBuilder.bindMarker("span_name"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("span_name", QueryBuilder.bindMarker("span_name"))));
 
     insertTraceIdByServiceName = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("service_name_index")
             .value("service_name", QueryBuilder.bindMarker("service_name"))
             .value("bucket", QueryBuilder.bindMarker("bucket"))
             .value("ts", QueryBuilder.bindMarker("ts"))
-            .value("trace_id", QueryBuilder.bindMarker("trace_id"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("trace_id", QueryBuilder.bindMarker("trace_id"))));
 
     insertTraceIdBySpanName = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("service_span_name_index")
             .value("service_span_name", QueryBuilder.bindMarker("service_span_name"))
             .value("ts", QueryBuilder.bindMarker("ts"))
-            .value("trace_id", QueryBuilder.bindMarker("trace_id"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("trace_id", QueryBuilder.bindMarker("trace_id"))));
 
     insertTraceIdByAnnotation = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("annotations_index")
             .value("annotation", QueryBuilder.bindMarker("annotation"))
             .value("bucket", QueryBuilder.bindMarker("bucket"))
             .value("ts", QueryBuilder.bindMarker("ts"))
-            .value("trace_id", QueryBuilder.bindMarker("trace_id"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("trace_id", QueryBuilder.bindMarker("trace_id"))));
 
     insertTraceIdBySpanDuration = session.prepare(
-        QueryBuilder
+        maybeUseTtl(QueryBuilder
             .insertInto("span_duration_index")
             .value("service_name", QueryBuilder.bindMarker("service_name"))
             .value("span_name", QueryBuilder.bindMarker("span_name"))
             .value("bucket", QueryBuilder.bindMarker("bucket"))
             .value("duration", QueryBuilder.bindMarker("duration"))
             .value("ts", QueryBuilder.bindMarker("ts"))
-            .value("trace_id", QueryBuilder.bindMarker("trace_id"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("trace_id", QueryBuilder.bindMarker("trace_id"))));
+  }
+
+  private RegularStatement maybeUseTtl(Insert value) {
+    return metadata.hasDefaultTtl
+        ? value
+        : value.using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_")));
   }
 
   /**
@@ -173,27 +175,24 @@ final class CassandraSpanConsumer implements GuavaSpanConsumer {
               span.id,
               span.annotations.hashCode(),
               span.binaryAnnotations.hashCode()),
-          ByteBuffer.wrap(Codec.THRIFT.writeSpan(span)),
-          spanTtl
-      ));
+          ByteBuffer.wrap(Codec.THRIFT.writeSpan(span))));
 
       for (String serviceName : span.serviceNames()) {
         // SpanStore.getServiceNames
-        futures.add(storeServiceName(serviceName, indexTtl));
+        futures.add(storeServiceName(serviceName));
         if (!span.name.isEmpty()) {
           // SpanStore.getSpanNames
-          futures.add(storeSpanName(serviceName, span.name, indexTtl));
+          futures.add(storeSpanName(serviceName, span.name));
         }
 
         if (span.timestamp != null) {
           // QueryRequest.serviceName
-          futures.add(storeTraceIdByServiceName(serviceName, span.timestamp,
-              span.traceId, indexTtl));
+          futures.add(storeTraceIdByServiceName(serviceName, span.timestamp, span.traceId));
 
           // QueryRequest.spanName
           if (!span.name.isEmpty()) {
             futures.add(storeTraceIdBySpanName(
-                serviceName, span.name, span.timestamp, span.traceId, indexTtl));
+                serviceName, span.name, span.timestamp, span.traceId));
           }
 
           // QueryRequest.min/maxDuration
@@ -201,10 +200,10 @@ final class CassandraSpanConsumer implements GuavaSpanConsumer {
             // Contract for Repository.storeTraceIdByDuration is to store the span twice, once with
             // the span name and another with empty string.
             futures.add(storeTraceIdByDuration(
-                serviceName, span.name, span.timestamp, span.duration, span.traceId, indexTtl));
+                serviceName, span.name, span.timestamp, span.duration, span.traceId));
             if (!span.name.isEmpty()) { // If span.name == "", this would be redundant
               futures.add(storeTraceIdByDuration(
-                  serviceName, "", span.timestamp, span.duration, span.traceId, indexTtl));
+                  serviceName, "", span.timestamp, span.duration, span.traceId));
             }
           }
         }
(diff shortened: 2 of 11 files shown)


zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -211,26 +220,30 @@ private long toCacheInterval(long ms) {
...
+                      indexTtl));
+          }
       }
     }
     return transform(Futures.allAsList(futures), TO_VOID);
   }
 
+  @Override
+  public void close() {
+    assert !session.isClosed();
```
adriancole: we've been using the pattern of don't close something unless you opened it. This didn't open the session, so it shouldn't close it, right?

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -89,13 +91,20 @@ private long toCacheInterval(long ms) {
     }
   };
 
-  CassandraSpanConsumer(Session session, int bucketCount, int spanTtl, int indexTtl) {
+  public CassandraSpanConsumer(Session session, int bucketCount, Integer spanTtl, Integer indexTtl)
+            throws IllegalStateException {
+
+    if (!Schema.isLatestSchema(session.getLoggedKeyspace(), session)) {
```
adriancole: due to various problems doing eager I/O in constructors (not just here, elsewhere, too) we've moved this logic to SessionFactory. With a package private constructor, we don't need to check here, if that worked.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -44,8 +44,10 @@
 import static zipkin.storage.cassandra.CassandraUtil.durationIndexBucket;
 import static zipkin.storage.cassandra.CassandraUtil.iso8601;
 
-final class CassandraSpanConsumer implements GuavaSpanConsumer {
+public final class CassandraSpanConsumer implements GuavaSpanConsumer, AutoCloseable {
```
adriancole: We've been using Closeable consistently (including the Component definition itself), but only on things that open resources. Ex in SessionFactory, we use Closer to close them (there's no AutoCloser, for example). This is particularly helpful in cassandra where there are multiple levels of things to close.

The contract of Closeable is more like what we are using it for, for example, we treat close as an idempotent command for components that do I/O. These objects are generally long lived, and outside of tests they aren't wrapped in try with resources, rather they are shared and registered with shutdown hooks etc.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
michaelsembwever: I can make the schema check lazy. The reason for having it here is the ability for CassandraSpanConsumer to gracefully fail when the schema is incompatible. 

That will become more obvious when i push the brave-spancollector-cassandra PR.

The constructor needs to be public, same reason as the class needs to be public.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
michaelsembwever: good point. will fix.

michaelsembwever:
The tests I've added require the `ccm` binary available on the command line.
Not sure how to deal with that.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -89,13 +91,20 @@ private long toCacheInterval(long ms) {
     }
   };
 
-  CassandraSpanConsumer(Session session, int bucketCount, int spanTtl, int indexTtl) {
+  public CassandraSpanConsumer(Session session, int bucketCount, Integer spanTtl, Integer indexTtl)
```
adriancole: Is there any alternative we can have vs making this public? I'm concerned because we've had to change this api in the past in the scala project, and the more narrow interface, the better..

For example, can we do something to make CassandraStorage.builder().build().guavaSpanConsumer() work?

I'd be happy to help change things so that we can keep this class internal. For example, if you need a "bring your own session" thing

michaelsembwever: Sure. But let's take that discussion once i've opened the brave-spancollector-cassandra PR, so it becomes consumer-driven-development.

adriancole: perfect idea!

michaelsembwever:
ccm tests are fixed.

zipkin-storage/cassandra/src/main/resources/cassandra-schema-cql3-upgrade-1.txt
```diff
@@ -0,0 +1,34 @@
+
+// upgrade cql to add the default_time_to_live option on each table
```
adriancole: is there a minimum cassandra version this works with that we should note?

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraSpanConsumerTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        CassandraSpanConsumer repo = new CassandraSpanConsumer(cluster.get().connect(keyspace), 10, null, null);
+
+        Endpoint ep = Endpoint.create("service", 127 << 24 | 1, 8080);
+
+        Queue<Future> q = new ArrayDeque<>();
+
+        for (int i = 0 ; i < 100; ++i) {
+
+            long micros = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis())
```
adriancole: fyi you can slice TestObject.LOTS_OF_SPANS to reduce clutter making spans

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraSpanConsumerTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @AfterClass
+    public static void afterClass() {
+        if (null != ccm) {
+            ccm.stop();
+            ccm.remove();
+        }
+    }
+
+    @Test
+    public void testSpanConsumer() throws IOException, InterruptedException, ExecutionException {
```
adriancole: what is this test doing? Is its purpose to test multiple addresses? concurrent ops?

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -89,13 +91,20 @@ private long toCacheInterval(long ms) {
     }
   };
 
-  CassandraSpanConsumer(Session session, int bucketCount, int spanTtl, int indexTtl) {
+  CassandraSpanConsumer(Session session, int bucketCount, Integer spanTtl, Integer indexTtl)
+            throws IllegalStateException {
+
+    if (!Schema.isLatestSchema(session.getLoggedKeyspace(), session)) {
```
adriancole: as far as I can tell, the changes to this file are around checking the schema here and making the ttls optional. The thing I'm curious about is this.. if the schema is not out of date, then it includes default ttls in the schema itself. In that case would the ttls ever be used intentionally?

adriancole:
Thanks for doing this, @michaelsembwever I just have a few questions so I can help support this once merged.

Is there some version-specificness behind the default ttl?
Should we deprecate the ttl options, remove them, or use them to construct the schema?
What's the cardinal interest of the CCM test? We've so far been doing portable tests.. once I understand what's of interest here, I can help move other things to re-use it etc post merge.

Once you're done, I'll do a formatting cleanup on this branch to make the intended change more obvious, update the README and builder javadocs accordingly, squashing into your commit.

Looks very interesting, and would significantly reduce the amount of storage operations made by default!

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
adriancole: Also, I think this check needs to be moved out to the lazy component. This type isn't usable except via that component, so it would tighten up the error handling.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
```diff
@@ -173,7 +179,17 @@ public CassandraStorage build() {
...
+      lazyAttemptTimestamp = System.currentTimeMillis();
+      try {
+        return new CassandraSpanConsumer(session.get(), bucketCount, spanTtl, indexTtl);
+      } catch (IllegalStateException ex) {
+          LOG.error("uninitialized CassandraSpanCollector due to out-of-date schema. no spans will be collected.", ex);
+      } finally {
+          LOG.info("initialized CassandraSpanCollector. now collecting spans.");
+      }
+    }
+    return null;
```
adriancole: returning null isn't something anyone would expect.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
```diff
@@ -173,7 +179,17 @@ public CassandraStorage build() {
   }
 
   @Override protected CassandraSpanConsumer computeGuavaSpanConsumer() {
-    return new CassandraSpanConsumer(session.get(), bucketCount, spanTtl, indexTtl);
+    if (lazyAttemptTimestamp + TimeUnit.SECONDS.toMillis(1) < System.currentTimeMillis()) {
+      lazyAttemptTimestamp = System.currentTimeMillis();
+      try {
```
adriancole: this check debilitates the component from a storage point of view. We currently don't have any mechanism to support a degraded (read-only) component. A side effect of doing the `isLatestSchema` check here as opposed to the LazySession component is that it makes the health check not useful. Ex the health check will pass, eventhough storage will fail.

adriancole:
One other thing I'm noticing is that this has a hard-break based on a schema.

One of our contracts about zipkin 1.0 is that it works with existing schema. There are many who install schema manually, for example, and we've been explicitly asked to keep v1.0 schema compatible.

If introduce a hard-break based on schema, we ought to do that post 1.0, similarly to how we've discussed not breaking the mysql schema until post 1.0.

If there's a way for us to work even when the upgrade fails (which most code seems to be ok with), then sounds fine to add this to 1.0.

FWIW, I think we can do 1.0 within the next week regardless, as there are no more blockers I'm aware of.

.travis.yml
```diff
@@ -30,6 +30,7 @@ before_install:
   - curl -SL http://downloads.datastax.com/community/dsc-cassandra-2.2.5-bin.tar.gz | tar xz
   - sed -i -e 's/INFO/WARN/g' dsc-cassandra-*/conf/logback.xml
   - sudo dsc-cassandra-*/bin/cassandra
+  - sudo pip install ccm
```
adriancole: ps right now, CassandraFixture skips tests when dependencies aren't present (ex ccm isn't installed). If we end up using ccm, we should probably remove the old way, so we don't need to skip in different ways. If using ccm is the best way, for example, I could help pull the "ccm" part into a separate pull request and help with the test fixture infrastructure around it.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
adriancole: yet another thing I notice while looking at impact of things.. this isn't guarded on `ensureSchema`. So, we should move it to the same place so that the guard applies consistently (ex. ensureSchema would add or update).

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
```diff
@@ -51,8 +53,8 @@ public static Builder builder() {
     String password;
     int maxTraceCols = 100000;
     int bucketCount = 10;
-    int spanTtl = (int) TimeUnit.DAYS.toSeconds(7);
-    int indexTtl = (int) TimeUnit.DAYS.toSeconds(3);
+    Integer spanTtl = (int) TimeUnit.DAYS.toSeconds(7);
```
adriancole: this ties into my earlier question, but should we still be defaulting these? I'm wondering if it wouldn't be cleaner to have a flag that disables setting ttl instead? For example, when we disable ttl, we ignore the values set here, as opposed to asking someone to set both of them to null.

```
builder.implicitTtl(); // use defaults as opposed to explicitly setting TTLs
```

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
```diff
@@ -173,7 +179,17 @@ public CassandraStorage build() {
   }
 
   @Override protected CassandraSpanConsumer computeGuavaSpanConsumer() {
-    return new CassandraSpanConsumer(session.get(), bucketCount, spanTtl, indexTtl);
+    if (lazyAttemptTimestamp + TimeUnit.SECONDS.toMillis(1) < System.currentTimeMillis()) {
```
adriancole: ps I still don't think it goes here, but if we are looking to do a "try once", we'd probably want to comment it, test it, and likely cache the exception vs logging. that would be more likely to not break things like healthcheck.

Problem with doing either trying once by logging or caching the exception is that we can't heal.

You have to reboot your process to heal. We've done quite a lot of work recently so that we can heal (ex survive race conditions and network glitches). For this reason, such a "try once" feature should probably be a separate pull request as it is likely to cause support issues later.

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraSpanConsumerTest.java
michaelsembwever: huh?

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraSpanConsumerTest.java
michaelsembwever: just the basic test that span consummation works.

zipkin-storage/cassandra/src/main/resources/cassandra-schema-cql3-upgrade-1.txt
michaelsembwever: no. default_time_to_live has been around for long enough.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
michaelsembwever: it is protected within the class and superclass.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
michaelsembwever: The idea is that this can heal without a reboot.

That's important when the consumer is running inside a client and restarting client applications due to an instrumentation library is not feasible.

adriancole: The code already handled this though. It only cached the result on
non-exception.

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
adriancole: The general comment is that the logic around saving off a timestamp and
explicitly returning null is untested and confusing (to me). One of us
should backfill tests on logic like this if it is really needed.

zipkin-storage/cassandra/src/main/resources/cassandra-schema-cql3-upgrade-1.txt
adriancole: Great

zipkin-storage/cassandra/src/test/java/zipkin/storage/cassandra/CassandraSpanConsumerTest.java
adriancole: OK then you can remove it as there are a dozen or more other tests that
exercise span consumption.

michaelsembwever: into a running cassandra?

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraStorage.java
michaelsembwever: It's not about caching successful results, but ensuring re-attempts (because of failures) to initialise only happen once per second.

This is because the cost of initialisation is a lot more than span consummation. A re-attempt to initialise every span will bring down client applications.

adriancole: ahah.. well there's been one benefit of spinning wheels which is that we
got to the essence of what you're looking for! In other words we got out of
this rut.

This problem will occur in several places in cassandra alone, and of course
apply to others as well. Let's top-level this as fixing in only one place
only helps the one occurrence we're seeing here. I will take this on.

adriancole:
extracted the wait on exception thing here after discussion with others #251

adriancole:
@michaelsembwever FYI just bumped this so that we test both with and without the latest schema. Going to sanity check by manual testing now.

zipkin-server/README.md
```diff
@@ -101,10 +101,8 @@ supports version 2.2+ and applies when `STORAGE_TYPE` is set to `cassandra`:
     * `CASSANDRA_CONTACT_POINTS`: Comma separated list of hosts / ip addresses part of Cassandra cluster. Defaults to localhost
     * `CASSANDRA_LOCAL_DC`: Name of the datacenter that will be considered "local" for latency load balancing. When unset, load-balancing is round-robin.
     * `CASSANDRA_MAX_CONNECTIONS`: Max pooled connections per datacenter-local host. Defaults to 8
-    * `CASSANDRA_ENSURE_SCHEMA`: Ensuring that schema exists, if enabled tries to execute script /zipkin-cassandra-core/resources/cassandra-schema-cql3.txt. Defaults to true
+    * `CASSANDRA_ENSURE_SCHEMA`: Ensuring cassandra has the latest schema. If enabled tries to execute scripts in the classpath prefixed with `cassandra-schema-cql3`. Defaults to true
     * `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD`: Cassandra authentication. Will throw an exception on startup if authentication fails. No default
-    * `CASSANDRA_SPAN_TTL`: Time-to-live in seconds for span data. Defaults to 604800 (7 days)
```
adriancole: intentionally un-documented these

adriancole:
log from a new server running against the old schema upgrading to the new schema. Note I posted from spigo to test that it worked.

```
2016-06-01 15:54:44.915  INFO 98416 --- [pool-1-thread-1] zipkin.storage.cassandra.Schema          : Upgrading schema /cassandra-schema-cql3.txt
```

adriancole:
Log when CASSANDRA_ENSURE_SCHEMA=false

```
2016-06-01 16:02:16.995  WARN 828 --- [nio-9411-exec-1] zipkin.storage.cassandra.Schema          : schema lacks default ttls: apply /cassandra-schema-cql3-upgrade-1.txt, or set CassandraStorage.ensureSchema=true
```

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
```diff
@@ -97,64 +100,62 @@ private long toCacheInterval(long ms) {
...
             .value("duration", QueryBuilder.bindMarker("duration"))
             .value("ts", QueryBuilder.bindMarker("ts"))
-            .value("trace_id", QueryBuilder.bindMarker("trace_id"))
-            .using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_"))));
+            .value("trace_id", QueryBuilder.bindMarker("trace_id"))));
+  }
+
+  private RegularStatement maybeUseTtl(Insert value) {
+    return metadata.hasDefaultTtl ? value
+        : value.using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_")));
```
michaelsembwever: nice approach.

trivial, but can i vote for conditional statements with ? and : each starting a new line?
eg

    return metadata.hasDefaultTtl 
            ? value
            : value.using(QueryBuilder.ttl(QueryBuilder.bindMarker("ttl_")));

michaelsembwever:
Nice. :+1: 

Like especially what's done with the tests.

michaelsembwever:
I can't get this to compile for me.
This line fails with `incompatible types: bad type in conditional expression`
https://github.com/openzipkin/zipkin-java/blob/master/zipkin-autoconfigure/storage-cassandra/src/main/java/zipkin/autoconfigure/storage/cassandra/ZipkinCassandraStorageAutoConfiguration.java#L43

adriancole:
Weird. Which JDK are you using?

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
adriancole: Sure thing!

michaelsembwever:
Am using 1.8.0_74
having some odd build problems with both zipkin-java and brave at the moment. leaving it for tomorrow.

adriancole:
OK.  You're using the maven wrapper, right? ./mvnw as opposed to mvn

michaelsembwever:
No :-(

zipkin-storage/cassandra/src/main/java/zipkin/storage/cassandra/CassandraSpanConsumer.java
adriancole: done

PR #242, opened 2016-05-20
Namespaces spring configuration to match package names

### LINKED ISSUE #239: Make AutoConfiguration tests
AutoConfiguration packages include property classes that map to configuration. Currently, the only way for us to find any bugs in the mapping is to manually set these and watch it work or fail.

This needs to be hardened before we go to version 1.0 as it is too easy to create config bugs.

adriancole:
This is a second go at namespacing spring configuration. This time, it
includes tests. I've also manually built docker to make sure we didn't
break it.

See #235 (for last attempt)
Fixes #239

### DIFF
--- a/zipkin-autoconfigure/collector-scribe/src/test/java/zipkin/autoconfigure/collector/scribe/ZipkinScribeCollectorAutoConfigurationTests.java
+++ b/zipkin-autoconfigure/collector-scribe/src/test/java/zipkin/autoconfigure/collector/scribe/ZipkinScribeCollectorAutoConfigurationTests.java
@@ -1,0 +1,97 @@
+/**
+ * Copyright 2015-2016 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package zipkin.autoconfigure.collector.scribe;
+
+import org.junit.After;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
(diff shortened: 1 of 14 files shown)


adriancole:
cc @shakuzen @marcingrzejszczak

zipkin-autoconfigure/collector-scribe/src/test/java/zipkin/autoconfigure/collector/scribe/ZipkinScribeCollectorAutoConfigurationTests.java
```diff
@@ -0,0 +1,109 @@
...
+    addEnvironment(context, "zipkin.collector.scribe.port", "9999");
+    context.register(PropertyPlaceholderAutoConfiguration.class,
+        ZipkinScribeCollectorAutoConfiguration.class, InMemoryConfiguration.class);
+    context.refresh();
+
+    assertThat(context.getBean(ZipkinScribeCollectorProperties.class).getPort())
+        .isEqualTo(9999);
+  }
+
+  static void addEnvironment(AnnotationConfigApplicationContext context, String... pairs) {
```
shakuzen: Why not use the `EnvironmentTestUtils` to do this? Like [this](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/PropertyPlaceholderAutoConfigurationTests.java#L52)

adriancole: If that's in some published jar, works for me.

shakuzen: Looks like it is included in the `spring-boot` jar, so we already have it available in this code without any dependency changes.

adriancole: Oh nice. Ill switch to this then

adriancole: done! Thanks for the recommendation

PR #246, opened 2016-05-23
Exposes the underlying client in use by the storage component

adriancole:
When integrating with other tools or testing, it is helpful to reuse
the same connection as zipkin does.


anuraaga:
LGTM - thanks!

PR #251, opened 2016-05-26
Teaches LazyCloseable to remember exceptions for a bit

### LINKED ISSUE #248: LazyCloseable could wait on exception
LazyCloseable (an internal class) is used for for network connections. The attempt to do something like creating a connection has a cost. It is even more expensive to try to do something like install a schema.

There was a comment on an otherwise unrelated issue that we should slow down attempts to connect to cassandra to [no more than once per second](https://github.com/openzipkin/zipkin-java/pull/240#discussion-diff-64503761).

In practice, there would only be one attempt per second if using brave because brave schedules span writes to a single thread, and that polling interval can be no shorter than one second. However, this "being saved by the bell" is something that only applies if using brave's base class, and wouldn't apply to normal collectors, which may have unbounded callers (who don't stand in a nice 1 second spaced line!).

Instead of being "saved by the bell", we could make either a global decision or a configuration option available to slow down tries.

For example, making lazy closeables try no more than once per second is a simple thing we could do, even by hard-coding at first. By hard-coding, we don't give flexibility, but we also don't have to bubble up a configuration option through the several components that would be impacted.

Thoughts?

adriancole:
Exceptions computing I/O intensive values can harm the process or the
backend. This teaches LazyCloseable to remember exceptions for a bit to
reduce that damage. The value is hard-set to 1 second which allows us
to mitigate the issue without adding a configuration burden.

Fixes #248

### DIFF
--- a/zipkin/src/main/java/zipkin/internal/Lazy.java
+++ b/zipkin/src/main/java/zipkin/internal/Lazy.java
@@ -34,10 +34,19 @@ public abstract class Lazy<T> {
       synchronized (this) {
         result = instance;
         if (result == null) {
-          instance = result = compute();
+          instance = result = tryCompute();
         }
       }
     }
     return result;
+  }
+
+  /**
+   * This is called in a synchronized block when the value to memorize hasn't yet been computed.
+   *
+   * <p>Extracted only for LazyCloseable, hence package protection.
+   */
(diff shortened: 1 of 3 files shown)


michaelsembwever:
+1

PR #257, opened 2016-05-31
Adds docs on how storage tests run

adriancole:
Some time back, we made sure developers aren't required to install and
run all data stores as a prerequisite of developing Zipkin. What we did
was made each of these conditional on environment, and skip integration
tests otherwise. We setup Travis such that all datastores are available,
so that no tests are skipped during pull requests.

What we didn't do was document what was going on, and this led to
confusion from different people who were curious about why pull requests
fail while local tests pass, or if integration tests run at all.

### DIFF
--- a/zipkin-storage/elasticsearch/README.md
+++ b/zipkin-storage/elasticsearch/README.md
@@ -13,3 +13,20 @@ to remove indices older than the point you are interested in.
 `zipkin.storage.elasticsearch.ElasticsearchStorage.Builder` includes defaults
 that will operate against a local Elasticsearch installation.
 
+## Testing this component
+This module conditionally runs integration tests against a local Elasticsearch instance.
+
+Tests are configured to automatically access Elasticsearch started with its defaults.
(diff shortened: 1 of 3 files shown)


adriancole:
cc @michaelsembwever @anuraaga @esbie

zipkin-storage/elasticsearch/README.md
```diff
@@ -13,3 +13,17 @@ to remove indices older than the point you are interested in.
...
+you'll notice tests are silently skipped.
+```
+Results :
+
+Tests run: 50, Failures: 0, Errors: 0, Skipped: 48
+```
+
+This behaviour is intentional: We don't want to burden developers with
+installing and running all storage options to test unrelated change.
+That said, all integration tests run on pull request via Travis.
```
anuraaga: Maybe also add

To actually run tests, download an elasticsearch 2.x distribution, extract it, and run bin/elasticsearch. Tests are configured to automatically access elasticsearch started with its defaults.

adriancole: done. thanks!

